PR #199, opened 2020-10-21
Problem: medium shiftleft scan findings (fix #127)

### LINKED ISSUE #127: Problem: medium shiftleft scan findings
some can be fixed, others may be suppressed in app.go (initRootCmd) and gentx.go (writeSignedGenTx):
1. "Expect file permissions to be 0600 or less." (https://cwe.mitre.org/data/definitions/276.html)
2. "Deferring unsafe method "Close" on type "*os.File"." (https://cwe.mitre.org/data/definitions/703.html)
3. "Potential file inclusion via variable." (https://cwe.mitre.org/data/definitions/22.html)

leejw51crypto:
Solution: change filemode to 600, check filepath validity, remove `defer` in file.close


cmd/chain-maind/app/app.go
```diff
@@ -142,26 +143,37 @@ func initRootCmd(rootCmd *cobra.Command, encodingConfig params.EncodingConfig) {
...
 			return err
 		}
 		if _, err := file.Seek(0, 0); err != nil {
+			file.Close()
 			return err
 		}
-		return json.NewEncoder(file).Encode(&genesis)
+
+		ret := json.NewEncoder(file).Encode(&genesis)
+		file.Close()
```
devashishdxt, line 175: Can we remove the earlier file.Close()?

leejw51crypto: previously, it was `defer flie.Close`, so closed after function returns.
now it explicitly close file

leejw51crypto:
i'll add unit test for the coverage

cmd/chain-maind/app/app.go
yihuang: defer can close the file in case of panic.

x/genutil/client/cli/gentx.go
```diff
@@ -244,19 +245,38 @@ func readUnsignedGenTxFile(clientCtx client.Context, r io.Reader) (sdk.Tx, error
...
-	outputFile, err := os.OpenFile(outputDocument, os.O_CREATE|os.O_EXCL|os.O_WRONLY, 0644)
+	outputFile, err := os.OpenFile(outputDocument, os.O_CREATE|os.O_EXCL|os.O_WRONLY, 0600)
+	if !IsValidPath(outputDocument) {
+		return fmt.Errorf("insecure filepath %s", outputDocument)
+	}
+
 	if err != nil {
 		return err
 	}
-	defer outputFile.Close()
```
yihuang, line 252: why remove this defer line, isn't defer is better than close manually?

leejw51crypto: i think so, too.
but code analysis tool recommends `not use defer to file.close` in issue #127

cmd/chain-maind/app/app.go
leejw51crypto: in golang, there is no exception,
so this is the same with previous version

x/genutil/client/cli/gentx.go
yihuang: https://www.joeshaw.org/dont-defer-close-on-writable-files/
I guess the objection to defer close method is it don't check and properly handle the return value of it, so if we are simply change to manually call the close method, we don't solve the real issue. and we lost the handling of panic.

cmd/chain-maind/app/app.go
yihuang: > in golang, there is no exception,
> so this is the same with previous version

golang has panic, and defer can run when that happens.

leejw51crypto: ok, i'll check

x/genutil/client/cli/gentx.go
leejw51crypto: ok, i'll modify based upon the link

tomtau: one other option: https://github.com/securego/gosec/issues/512#issuecomment-675286833

yihuang: we can make a extension method `file.CloseOrWarn()` ? ;D

cmd/chain-maind/app/app.go
```diff
@@ -142,26 +143,37 @@ func initRootCmd(rootCmd *cobra.Command, encodingConfig params.EncodingConfig) {
 		config.SetRoot(clientCtx.HomeDir)
 		path := config.GenesisFile()
 
-		file, err := os.OpenFile(path, os.O_RDWR, 0644)
+		file, err := os.OpenFile(path, os.O_RDWR, 0600)
+		if !chaingenutilcli.IsValidPath(path) {
+			return fmt.Errorf("insecure filepath %s", path)
+		}
```
tomtau, line 149: For this file path finding, two things:
1. it's likely a false positive, as the scope of this is very limited -- at least in the case of genutil
2. this "fix" is a bit problematic:
- it's done after opening the file
- it does some custom input validation with a deny list

So I suggest keeping this simple and either:
- suppress this warning (`// nolint: gosec`)
- or just do a simple path cleanup / canonical path -- `filepath.Clean(path)`

tomtau (changes requested):
- defer for closing files is still OK -- it's a common pattern -- the problematic part is missing the error code -- the easiest way may be just to defer a function that will close the file, but also log any errors (if they happen

- for the path validation, keeping it simple with filepath.Clean

x/genutil/client/cli/gentx.go
leejw51crypto: i'll check `CloseOrWarn`

cmd/chain-maind/app/app.go
```diff
@@ -142,11 +143,20 @@ func initRootCmd(rootCmd *cobra.Command, encodingConfig params.EncodingConfig) {
...
 		if err != nil {
 			return err
 		}
-		defer file.Close()
+
+		defer func() {
+			cerr := file.Close()
+			if err == nil {
+				err = cerr
+			}
```
tomtau, line 157: and if err is not nil, it'll be lost? I'm not sure of the exact defer semantics, but perhaps logging the error (if cerr is not nil) won't hurt?

yihuang: I think assign to err at here don't have any effect, it's in panic handling process here, the err is not consumed by anybody.

leejw51crypto: yes, i changed to display the log

tomtau (approved):

PR #211, opened 2020-10-23
Problem (fix #209): build chain-maind with nix-build too slow for local dev

### LINKED ISSUE #209: Problem: build chain-maind with nix-build too slow for local dev
running `vend` locally takes forever, especially when one want to debug with local cosmos-sdk directory, you need to run that   for each time you modify cosmos-sdk code.
so better to disconnect pystarport and chain-maind when test locally, install chain-maind with `go install`, pystarport find it through `PATH`.

yihuang:
Solution:
- find chain-maind on `PATH` when dev locally
- don't collect coverage report when test locally

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -143,8 +143,8 @@ clean-docker-compose: localnet-stop
 ###############################################################################
 # nix installation: https://nixos.org/download.html
 nix-integration-test: check-network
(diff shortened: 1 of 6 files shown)


leejw51crypto (approved):
lgtm

tomtau (approved):

Issue #194, opened 2020-10-19
Problem: missing grpc integration test


leejw51crypto:
add grpc integrataion test
1. make tx
2. sign tx
3. broadcast with grpc

leejw51crypto:
started

leejw51crypto:
compile python communication code using `grpc protoc plugin`,
get validators working fine

now trying to make tx, and broadcast through tendermint grpc port

leejw51crypto:
grpc query working fine with protobuf compiler + grpc protoc plugin(for python)
and doing tx broadcasting via tendermint grpc port

leejw51crypto:
tx broadcasting working,
now adding integration test

leejw51crypto:
will submit validators test first as WIP

PR #236, opened 2020-11-02
Problem: missing grpc integration test (wip #194)

leejw51crypto:
add grpc integration test for get validators


tomtau:
some small py linting changes: https://github.com/crypto-com/chain-main/pull/236/checks?check_run_id=1340772489#step:5:5738

integration_tests/shell.nix
```diff
@@ -24,4 +24,8 @@ mkShell {
       nixpkgs-fmt
       cosmovisor
     ];
+  shellHook = ''
+    # prefer local pystarport directory for development
+    export PYTHONPATH=$PWD/pystarport:$PWD/pystarport/pystarport/proto_python:$PYTHONPATH
```
yihuang, line 29: I think you don't need this for ci testing

yihuang (commented):
- `proto_python` can be installed with pystarport package, no need to set extra `PYTHONPATH`. 
- can you also add the commands somewhere to update these generated files?

leejw51crypto:
> * `proto_python` can be installed with pystarport package, no need to set extra `PYTHONPATH`.
> * can you also add the commands somewhere to update these generated files?

ok,i'll check PYTHONPATH,
for converting, i'll add markdown to update these files

leejw51crypto:
i'm fixing linting issues,
i'll exclude autogenerated files

pystarport/pystarport/proto_python/confio/proofs_pb2_grpc.py
```diff
@@ -0,0 +1,3 @@
+# Generated by the gRPC Python protocol compiler plugin. DO NOT EDIT!
```
tomtau, line 1: do we want to commit these protoc generated files for tests instead of generating them on the fly?

tomtau (changes requested):
merge conflict with latest master

integration_tests/shell.nix
leejw51crypto: other autogenerated python code need to `proto_python` path as a module,
also to run python module elsewhere, it's better to specify absolute path in shell.nix

pystarport/pystarport/proto_python/confio/proofs_pb2_grpc.py
leejw51crypto: i'll add grpc cli python utilities afterwards.
we can add eventcall via grpc, and other useful features as python codes in pystarport.

leejw51crypto:
rebased the master, adding regenerateing proto script

integration_tests/shell.nix
yihuang: I guess you can config import path to the tools?

leejw51crypto: programatically adding path?

leejw51crypto: i'll check protoc option

tomtau:
some nix syntax error: https://github.com/crypto-com/chain-main/pull/236/checks?check_run_id=1345737471#step:5:6

leejw51crypto:
integration test on local is fine, but breaks on github
i'm inspecting now

leejw51crypto:
fixing module issue.
i'm checking two solutions
1. make another python module for proto_python
2. add sub module inside pystarport, re-generate python codes with prefix

leejw51crypto:
moved protobuf files as sub-module of pystarport

leejw51crypto:
protoc export doesn't support prefix, so used __init__.py
now writing protoc output generating script.

leejw51crypto:
rebased, added converter

tomtau:
I guess this is superseded by https://github.com/crypto-com/chain-main/pull/249 ?

PR #248, opened 2020-11-05
Problem: (fix #135) hw wallet not tested on windows

### LINKED ISSUE #135: Problem: hw wallet not tested on windows
check hw-wallet(ledger) is working fine on windows
also make sure that integration test working fine on windows

linfeng-crypto:
* test ledger device manually on windows with the binary build by the goreleaser , it works fine.
* integration test can not run on windows for the following reason:
    * nix does not support the windows platform
    * supervisor does not work fine on windows in my test
* Currently, [WSL does not have USB pass-through access](https://github.com/microsoft/WSL/issues/5158), so can't use ledger device on WSL.
* integration test works fine on windows WSL 2, if get an error like `docker-credential-desktop.exe not installed or not available in PATH`,  see [here](https://github.com/docker/for-mac/issues/3785)
* can not expose the container's port outside using the host network mode on WSL 2, so change to container link mode.

### DIFF
--- a/pystarport/pystarport/cluster.py
+++ b/pystarport/pystarport/cluster.py
@@ -30,6 +30,7 @@ from .utils import build_cli_args_safe, format_doc_string, interact, write_ini
 CHAIN = "chain-maind"  # edit by nix-build
 ZEMU_HOST = "127.0.0.1"
 ZEMU_BUTTON_PORT = 9997
+ZEMU_GRPC_SERVER_PORT = 3002
 # dockerfile is integration_test/hardware_wallet/Dockerfile
 ZEMU_IMAGE = "cryptocom/builder-zemu:latest"
 IMAGE = "docker.pkg.github.com/crypto-com/chain-main/chain-main-pystarport:latest"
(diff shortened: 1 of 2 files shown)


pystarport/pystarport/cluster.py
```diff
@@ -79,11 +79,23 @@ def __init__(self):
         self.containers = []
 
     def start(self):
-        for (name, cmd) in self.cmds.items():
+        host_config_ledger = self.client.api.create_host_config(
+            auto_remove=True, port_bindings={9997: 9997, 3002: 3002}
```
leejw51crypto, line 83: how about make const variable, declare to the top

pystarport/pystarport/cluster.py
```diff
@@ -79,11 +79,23 @@ def __init__(self):
...
+        )
+        container_ledger = self.client.api.create_container(
+            ZEMU_IMAGE,
+            self.cmds[self.ledger_name],
+            name=self.ledger_name,
+            ports=[9997, 3002],
```
leejw51crypto, line 89: using const variable?

leejw51crypto (approved):
lgtm

pystarport/pystarport/cluster.py
linfeng-crypto: sure

PR #249, opened 2020-11-05
Problem: missing grpc integration test (fix #194)

### LINKED ISSUE #194: Problem: missing grpc integration test

leejw51crypto:
Solution: add query, tx broadcast integration test using grpc


pystarport/pystarport/convert.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
tomtau, line 1: could this be executed on the fly?

tomtau (commented):
looks ok, but before merging this, probably good to consider whether to: 1) commit these 50,000 LOC of Python + 180+ files to the repo, or 2) generate this protobuf boilerplate on the fly before test execution (if that's possible)? @yihuang @calvinaco @devashishdxt @linfeng-crypto

pystarport/pystarport/convert.sh
leejw51crypto: protoc can be installed in nix
grpc plugin should be compiled,
i'll check  grpc plugin in nix

leejw51crypto:
submit again

PR #253, opened 2020-11-08
Problem: missing grpc integration test (fix #194)

### LINKED ISSUE #194: Problem: missing grpc integration test

leejw51crypto:
Solution: add grpc query, tx broadcast integration test
generate proto+grpc code on the fly

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,3 +1,4 @@
 [flake8]
 max-line-length = 88
 extend-ignore = E203
+exclude = .git,__pycache__,./pystarport/pystarport/cosmos-sdk,./pystarport/pystarport/tendermint,./pystarport/pystarport/proto_python
--- a/pystarport/pystarport/convert.sh
+++ b/pystarport/pystarport/convert.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+OUTPUT=./proto_python
+COSMOS=./cosmos-sdk
+TENDERMINT=./tendermint
+TMP=$(whereis grpc_python_plugin)
+PLUGIN="$(cut -d' ' -f2 <<<"$TMP")"
+mkdir $OUTPUT
+git clone --branch v0.40.0-rc2 https://github.com/cosmos/cosmos-sdk.git
+git clone --branch v0.34.0-rc5 https://github.com/tendermint/tendermint.git
+cp -Rf $COSMOS/third_party/proto/* $COSMOS/proto/ 
+# cosmos
+protoc --proto_path=$COSMOS/proto --proto_path=$COSMOS/third_party/proto --python_out=$OUTPUT $(find $COSMOS/proto/cosmos -iname "*.proto") --grpc_python_out=$OUTPUT  --plugin=protoc-gen-grpc_python=$PLUGIN
+# cosmos third-party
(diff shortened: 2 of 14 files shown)


leejw51crypto:
fixing 
- nix file lint issue
- grpc_python_plugin absolute path issue

pystarport/pystarport/convert.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+OUTPUT=./proto_python
+COSMOS=./cosmos-sdk
+TENDERMINT=./tendermint
+TMP=$(whereis grpc_python_plugin)
+PLUGIN="$(cut -d' ' -f2 <<<"$TMP")"
+mkdir $OUTPUT
+git clone https://github.com/cosmos/cosmos-sdk.git
+git clone https://github.com/tendermint/tendermint.git
```
tomtau, line 9: not sure whether to pin this against master (which can be non-reproducible, but can possibly find out issues earlier) or against specific commits/tags (e.g. v0.40-rc2 for SDK) to match our code

tomtau: `git checkout <tag>`

leejw51crypto:
fixed lint issue, and integration test

tomtau (commented):
looks ok -- just one comment whether to pin the makeproto script against specific versions

pystarport/pystarport/convert.sh
leejw51crypto: ok, i'll pin the version

pystarport/pystarport/default.nix
```diff
@@ -0,0 +1,12 @@
+{ pkgs ? import ../../nix { }, commit ? "" }:
```
yihuang, line 1: - this is the directory of python package, better to put config files outside
- maybe we make the proto generated files as a nix package, which can be depend by pystarport package?

.flake8
```diff
@@ -1,3 +1,4 @@
 [flake8]
 max-line-length = 88
 extend-ignore = E203
+exclude = .git,__pycache__,./pystarport/pystarport
```
yihuang, line 4: you excluded the whole pystarport package from python linter?

pystarport/pystarport/convert.sh
tomtau: or `git clone -b v0.40-rc2  https://github.com/cosmos/cosmos-sdk.git` && `git clone -b v0.34-rc5  https://github.com/tendermint/tendermint.git`

.flake8
leejw51crypto: i'll change it, exclude only cosmos, tendermint

pystarport/pystarport/default.nix
leejw51crypto: i'll try not to use ../../nix

pystarport/pystarport/convert.sh
leejw51crypto: ok, looking good

yihuang: or `curl ...archive/hash.tar.gz`

tomtau (approved):

PR #258, opened 2020-11-10
Problem: no distinct options for liquid vs total supply retrieval

### LINKED ISSUE #78: Problem: no distinct options for liquid vs total supply retrieval
Currently, the "supply" will return the total supply -- @yihuang ? It'll also be good to be able to retrieve "circulating" supply (total supply - (total bonded + community pool + unvested amounts))

devashishdxt:
Solution: Created supply module to retrieve total and liquid supply.
Fixes #78


go.sum
```diff
@@ -115,6 +116,7 @@ github.com/cosmos/cosmos-sdk v0.34.4-0.20201022224044-5bebf2bd39cd h1:5UiE2afJO8
 github.com/cosmos/cosmos-sdk v0.34.4-0.20201022224044-5bebf2bd39cd/go.mod h1:KvZzKUEc7KrMiiIeU0zKPMMwuxxn+R0Leo8vBSDgOeQ=
 github.com/cosmos/cosmos-sdk v0.34.4-0.20201028155748-1b0d6540d8d5 h1:ALPZHfvrFRMyFDhGi3OTzHo7DUN+VENVNIwAnWOqHwQ=
 github.com/cosmos/cosmos-sdk v0.34.4-0.20201028155748-1b0d6540d8d5/go.mod h1:KvZzKUEc7KrMiiIeU0zKPMMwuxxn+R0Leo8vBSDgOeQ=
+github.com/cosmos/cosmos-sdk v0.39.1 h1:vhjf9PZh9ph8btAj9aBpHoVITgVVjNBpM3x5Gl/Vwac=
```
tomtau, line 119: these go.sum / go.mod updates may not be necessary?

proto/cosmos/base/v1beta1/coin.proto
```diff
@@ -0,0 +1,40 @@
+syntax = "proto3";
```
tomtau, line 1: for these third party protos, have a third_party dir, as in: https://github.com/cosmos/cosmos-sdk/tree/master/third_party ?

proto/supply/v1beta/genesis.proto
```diff
@@ -0,0 +1,18 @@
...
+package chainmain.supply.v1beta1;
+
+import "gogoproto/gogo.proto";
+
+option go_package = "github.com/crypto-com/chain-main/x/supply/types";
+
+// VestingAccounts stored in keeper
+message VestingAccounts {
+    // addresses defines addresses of all the vesting accounts at genesis
+    repeated string addresses = 1;
```
tomtau, line 11: does this need a proto definition / duplicate information to what's in genesis? why not just iterate accounts in genesis and check the account type?

x/supply/keeper/querier.go
```diff
@@ -0,0 +1,56 @@
...
+
+func queryTotalSupply(
+	ctx sdk.Context,
+	req abci.RequestQuery,
+	k Keeper,
+	legacyQuerierCdc *codec.LegacyAmino,
+) ([]byte, error) {
+	totalSupply := k.GetTotalSupply(ctx)
+
+	res, err := legacyQuerierCdc.MarshalJSON(totalSupply)
```
tomtau, line 34: is the legacy api worth adding if it's already exposed via grpc?

tomtau (commented):
it's ok; just perhaps some code can be reduced/simplified.
a small nit: the old SDK module used to be called "supply", so perhaps a different name for this ("supply-track" or something like that) in order not to confuse? but ok to keep the name

proto/cosmos/base/v1beta1/coin.proto
devashishdxt: Done

proto/supply/v1beta/genesis.proto
devashishdxt: I haven't seen any examples of directly reading whole genesis file in a modules `InitChain`. Instead the common pattern in `cosmos-sdk` is to have separate `app_state` for each module in `genesis.json` for initialisation. I've added code to inject vesting accounts in `genesis.json` in `add-genesis-account` command (similar to what we do for `bank` and `auth` modules).

go.sum
devashishdxt: Updated both files by running `go mod tidy`.

proto/supply/v1beta/genesis.proto
leejw51crypto: how about getting vesting information through keeper?
not directly refering to original genesis.json(of tendermint)

x/chainmain/client/cli/genaccounts.go
```diff
@@ -110,8 +111,10 @@ contain valid denominations. Accounts may optionally be supplied with vesting pa
 
 			balances := banktypes.Balance{Address: addr.String(), Coins: coins.Sort()}
 			baseAccount := authtypes.NewBaseAccount(addr, nil, 0, 0)
+			var vestingAddress string
 
 			if !vestingAmt.IsZero() {
```
leejw51crypto, line 116: vestingAmount?

proto/supply/v1beta/genesis.proto
devashishdxt: @leejw51crypto I don't think we can do that. For that, we'll have to loop through all the accounts in the system.

x/chainmain/client/cli/genaccounts.go
devashishdxt: You mean rename `vestingAddress` to `vestingAmount`?

leejw51crypto: renaming vestingAmt -> vestingAmount

devashishdxt: Yeah. Maybe we can do that in a separate PR as this one does not change that part of the code.

proto/supply/v1beta/genesis.proto
leejw51crypto: how about in InitGenesis
use this api of AccountKeeper
iterate all accounts , check vesting, append ?

// GetAllAccounts returns all accounts in the accountKeeper.
func (ak AccountKeeper) GetAllAccounts(ctx sdk.Context) (accounts []types.AccountI) {
	ak.IterateAccounts(ctx, func(acc types.AccountI) (stop bool) {
		accounts = append(accounts, acc)
		return false
	})

	return accounts
}

devashishdxt: Won't this iterate over all the accounts? We only want to iterate over accounts which were specified in `genesis.json`.

leejw51crypto: yes, but in genesis init part,
accounts are a few numbers, so i think it will be very fast

tomtau: genesis.json can be large-ish if the state is exported from a previous network -- e.g. in cosmohub, it's several MBs -- but since "initgenesis" is only executed once, it's probably ok to iterate over the accounts.

So I suggest either:
1. drop the duplicate data and just compute it once from genesis
2. or add extra validations (https://github.com/crypto-com/chain-main/pull/258/files#diff-f995c0703ad210807c8ed69e34026fbcaa1cfe5a0acb4d9f28cafdb9c6d98254R23) to make sure there's no mismatch between genesis accounts and this duplicate supply accounts.

maybe 1. is simpler?

leejw51crypto: option 1 seems good

devashishdxt: Done.

leejw51crypto (approved):
lgtm

tomtau (changes requested):
maybe a unit test or integration test to demonstrate it works?

devashishdxt:
> maybe a unit test or integration test to demonstrate it works?

Done

tomtau (approved):

PR #259, opened 2020-11-10
Temp PR. Just to check nix build on github actions.

devashishdxt:


x/supply/keeper/keeper.go
```diff
@@ -0,0 +1,127 @@
...
+	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+	distrtypes "github.com/cosmos/cosmos-sdk/x/distribution/types"
+	govtypes "github.com/cosmos/cosmos-sdk/x/gov/types"
+	minttypes "github.com/cosmos/cosmos-sdk/x/mint/types"
+	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
+)
+
+var (
+	// ModuleAccounts defines the module accounts which will be queried to get liquid supply
+	ModuleAccounts = []string{
```
yihuang, line 18: I think one can also sum these balances outside of the chain with hard coded module account list and the vested account list in genesis file?

Issue #219, opened 2020-10-27
Problem: missing python wallet and tx signing framework


leejw51crypto:
cosmospy package is too new, not compatible with ours.

we can write small secp256k1 framework for chain-maind.
features 

1. hdwallet 
2. tx signing
3. tx broadcasting

so that, this package can be used for integration test,too.

linfeng-crypto:
doing it now.

tomtau:
in chainlibpy

PR #281, opened 2020-11-23
Problem fix(#219): missing python wallet and tx signing framework

linfeng-crypto:
solution: add an integration test using cryptopy to sign transaction offline


tomtau (approved):
seems ok -- we may later need to rename cryptopy, as it collides with other package on pypy

PR #297, opened 2020-12-15
Problem: latest SDK release candidate is not used

tomtau:
Solution:
- switched code to 0.40.0-rc5
- based on https://github.com/crypto-com/chain-main/pull/275 to use denom conversion utils from Cosmos SDK


default.nix
```diff
@@ -99,45 +99,40 @@ rec {
...
       chain-maind-zemu.instrumented
-      tests
     ] ++ common-env;
   };
 
   # main entrypoint script to run integration tests
   run-integration-tests = pkgs.writeShellScriptBin "run-integration-tests" ''
-    export PATH=${ci-env-inner}/bin:$PATH
-    export TESTS=${ci-env-inner}/share/integration_tests
+    export PATH=${ci-env}/bin:$PATH
```
yihuang, line 122: I think we need a “set -e” here, to stop on error.

tomtau: yes, it's reporting now! thanks

app/app.go
```diff
@@ -81,46 +83,36 @@ import (
...
+	DefaultNodeHome string
 
 	// ModuleBasics defines the module BasicManager is in charge of setting up basic,
 	// non-dependant module elements, such as codec registration
 	// and genesis verification.
 	ModuleBasics = module.NewBasicManager(
 		auth.AppModuleBasic{},
 		genutil.AppModuleBasic{},
-		chainbank.NewAppModuleBasic(DefaultCoinParser),
+		bank.AppModuleBasic{},
```
devashishdxt, line 109: Coin parsing will be done by upstream?

cmd/chain-maind/app/app.go
```diff
@@ -254,6 +248,7 @@ func txCommand() *cobra.Command {
 		authcmd.GetEncodeCommand(),
 		authcmd.GetDecodeCommand(),
 		flags.LineBreak,
+		vestingcli.GetTxCmd(),
```
devashishdxt, line 251: Do we need to add this here?

go.sum
```diff
@@ -113,17 +113,26 @@ github.com/cosmos/cosmos-sdk v0.34.4-0.20201022224044-5bebf2bd39cd/go.mod h1:KvZ
 github.com/cosmos/cosmos-sdk v0.34.4-0.20201028155748-1b0d6540d8d5 h1:ALPZHfvrFRMyFDhGi3OTzHo7DUN+VENVNIwAnWOqHwQ=
 github.com/cosmos/cosmos-sdk v0.34.4-0.20201028155748-1b0d6540d8d5/go.mod h1:KvZzKUEc7KrMiiIeU0zKPMMwuxxn+R0Leo8vBSDgOeQ=
 github.com/cosmos/cosmos-sdk v0.39.1 h1:vhjf9PZh9ph8btAj9aBpHoVITgVVjNBpM3x5Gl/Vwac=
+github.com/cosmos/cosmos-sdk v0.39.2 h1:nLfCJMkUuFt7ansi/YvCxwwxLFrgHCA3cYP4sJKYQdk=
```
devashishdxt, line 116: Can we remove older versions of cosmos-sdk?

app/app.go
tomtau: yes, it's been merged upstream

cmd/chain-maind/app/app.go
tomtau: it's possible to create vesting accounts dynamically afaik -- so this adds client side commands for it

app/app.go
tomtau: it's called as `sdk.ParseCoinsNormalized` -- https://github.com/crypto-com/chain-main/pull/275

go.sum
tomtau: `go.sum` is a sort of append-only AFAIK -- not sure of clean up options @calvinaco

cmd/chain-maind/app/app.go
calvinaco: cc @lezzokafka With this we may be able to create vesting accounts after genesis is created.

go.sum
calvinaco: you can try to run `go mod tidy` and it will look for unused modules to remove

calvinaco (approved):
LGTM

cmd/chain-maind/app/app.go
devashishdxt: Just a side note: supply module calculations for liquid supply will not work for vesting accounts created after genesis.

go.sum
tomtau: thanks, `go mod tidy` removed some of those

PR #306, opened 2020-12-23
Problem: (fix #305) pystarport missing optional pubkey in config

### LINKED ISSUE #305: Problem: pystarport missing optional pubkey in config
When using pystarport to generate new genesis, one can specify consensus_key `pub` and `priv` in config.
It would be great to add pubkey as well. As some may use tmkms for initial validator, they are not using `priv_validator_key.json`

allthatjazzleo:
Solution: add option for pubkey in pystarport


examples/testnet.yaml
```diff
@@ -7,5 +7,7 @@ testnet-croeseid-1:
       # consensus_key:
       #   priv:
       #   pub:
+      # optional, using existing pubkey instead of using priv_validator_key.json, generated by tmkms
+      # pubkey: tcrocnclconspub1.....
```
yihuang, line 11: It already supports setting "consensus_key", the "priv_validator_key.json" will be overwritten by that. But you need to put private key in that.

yihuang (commented):
I think the use case already covered by the "consensus_key" feature.

PR #323, opened 2020-12-30
Problem (fix #294): no spec for subscriptions module

### LINKED ISSUE #294: Problem: no spec for subscriptions module
**Is your feature request related to a problem? Please describe.**
Not being able to have regular payments to subscribe for services.

**Describe the solution you'd like**
Can be something similar to https://eips.ethereum.org/EIPS/eip-1337
The spec should follow SDK `x/*/spec` format: https://github.com/cosmos/cosmos-sdk/tree/master/x/bank/spec

**Describe alternatives you've considered**
- having it done on the client-side only -- just "wallet UX" will hide the fact it uses regular transfers
- having a trusted third party / payment processor + 2-of-3 threshold on signing

**Additional context**
subscriptions outside of the blockchain context: https://developer.paypal.com/docs/api/subscriptions/v1

yihuang:
Solution:
- add one

TODO: change to pseudocode to Golang after impl.

### DIFF
--- a/x/subscription/spec/01_concepts.md
+++ b/x/subscription/spec/01_concepts.md
@@ -1,0 +1,26 @@
+<!--
+order: 1
+-->
+
+# Concepts
+
+## Subscription
+
+- Subscription plan owners create/stop plans on Chain, the plan prices are defined for specified token denominations, subscribers should do necessary token conversions in other ways, e.g. exchange.
+- Users subscribe/unsubscribe to plan.
+- Payments are collected automatically at the start of their specified intervals, the results of collections are written into block events.
+- Create plan transaction will consume a gas fee to cover the computational cost of future automatic payment collections.
+- If the automatic collection mechanism fails on some subscribers, Chain won't automatically retry later, the corresponding plan owner can either configure their plan to automatically cancel these subscribers or manually retry failed collections later.
+
+### Intervals
+
+Intervals are specified using [crontab syntax](https://crontab.guru/), for example:
+
+- `* * * * *`: At *every minute*
+- `0 8 * * *`: At *08**:**00*
+- `0 8 1 * *`: At *08**:**00* *on day-of-month 1*
+- `0 8 * * 1`: At *08**:**00* *on Monday*
+
+Convenient shortcuts:
+
+- `@monthly`: `0 0 * * *`
--- a/x/subscription/spec/02_state.md
+++ b/x/subscription/spec/02_state.md
@@ -1,0 +1,87 @@
+<!--
+order: 2
+-->
+
+# State
+
+The `x/subscription` stores the plans and subscriptions on chain:
+
+```golang
+type SubscriptionPlan struct {
+  id int,  // auto-increasing unique identifier
+  title string,
+  description string,
+  owner Address,  // beneficial owner of the plan
+  price Coins,  // price to pay for each period, Coins contains both amount and denomination
+  subscription_duration int,  // duration of subscriptions
+  cron_spec CronSpec,  // Configure time intervals, parsed from crontab syntax
+  cron_tz Timezone,  // timezone for cron_spec
+}
+
+type Subscription struct {
+  plan_id int,
+  subscriber Address,
+  create_time int, // the block time when subscription was created
+  // the timestamp of last successful collection,
+  // default to the current block time round-down against cron spec
+  // subscribers don't pay for the period it gets created in
+  last_collected_time int,
+  payment_failures int,  // times of failed payment collection
+}
+
+type GenesisState struct {
+  subscription_plans [SubscriptionPlan];
+  subscriptions [Subscription];
+}
+
+func (s *Subscription) next_collection_time() int {
+  var plan = get_plan(s.plan_id)
+  return round_up_time(s.last_collected_time, plan.cron_spec, plan.cron_tz)
+}
+
+func (s *Subscription) expiration_time() int {
+  var plan = get_plan(s.plan_id)
+  return s.create_time + plan.subscription_duration
+}
+
+// Parsed crontab syntax
+type CronSpec struct {
+  minute [CronValue]
+  hour [CronValue]
+  day [CronValue]
+  month [CronValue]
+  wday [CronValue]
+}
+CronValue = Any | Range(start, end, step) | Value(v)
+```
+
+## Round time
+
+We define two functions to round timestamp to the boundary of periods:
+
+```python
+def round_down_time(timestamp, cron_spec, cron_tx):
+    # return the largest timestamp which matches cron_spec and less or equal than timestamp
+
+def round_up_time(timestamp, cron_spec, cron_tx):
+    # return the smallest timestamp which matches cron_spec and greater than timestamp
+
+def count_period(begin_time, end_time, cron_spec, cron_tz):
+    # return the number of periods between two timestamps
+    count = 0
+    while True:
+      begin_time = round_up_time(begin_time, cron_spec, cron_tz)
+      if begin_time >= end_time:
+        break
+      count += 1
+    return count
+```
+
+To keep payment collection idempotent (no duplicated collection in same collection period), we record the last round-down timestamp of collection:
+
+```python
+period_index = round_down_time(block_time, cron_spec)
+if period_index > last_period_index:
+    # do the collection
+    last_period_index = period_index
+```
--- a/x/subscription/spec/03_messages.md
+++ b/x/subscription/spec/03_messages.md
@@ -1,0 +1,56 @@
+<!--
+order: 3
+-->
+
+# Messages
+
+## MsgCreateSubscriptionPlan
+
+Plan owners create plans, the message signer is the plan owner.
+
+```protobuf
+message MsgCreateSubscriptionPlan {
+  string title;
+  string description;
+  Coins price,  // amount + denomination, amount + denomination, ...
+  int32 subscription_duration,
+  repeated int32 collection_timestamps,
+}
+```
+
+## MsgStopSubscriptionPlan
+
+Sent by plan owners, also removes all corresponding subscriptions.
+
+```protobuf
+message MsgStopSubscriptionPlan {
+  int32 plan_id,
+}
+```
+
+## MsgCreateSubscription
+
+The message signer subscribe to the plan.
+
+```protobuf
+message MsgCreateSubscription {
+  int32 plan_id,
+}
+```
+
+It'll consume some gases for each collection period:
+
+```python
+ConsumeGas( count_period(create_time, expiration_time, plan.cron_spec, plan.cron_tz) * GasPerCollection )
+```
+
+## MsgStopSubscription
+
+Both the subscriber and the plan owner can stop a subscription.
+
+```protobuf
+message MsgStopSubscription {
+  int32 plan_id,
+  Address subscriber,
+}
+```
--- a/x/subscription/spec/06_begin_block.md
+++ b/x/subscription/spec/06_begin_block.md
@@ -1,0 +1,47 @@
+<!--
+order: 6
+-->
+
+# BeginBlock
+
+## Remove expired subscriptions
+
+```python
+for plan_id, subscriber, expiration_time in get_subscriptions_sorted_by_expiration_time():
+    if block_time >= expiration_time:
+        stop_subscription(plan_id, subscriber)
+    else:
+        break
+```
+
+## Collect payments
+
+`x/subscription` module automatically collect payments from subscribers at begin block, and write collection results into events.
+
+```python
+for subscription in get_subscriptions_sorted_by_next_collection_time():
+    plan = get_plan(subscription.plan_id)
+    if block_time >= subscription.next_collection_time():
+        current_period = round_down_time(block_time, plan.cron_spec, plan.cron_tz)
+        if current_period > subscription.last_collected_time:
+            if transfer(subscription.subscriber, plan.owner, plan.price):
+                subscription.payment_failures = 0
+                emit Event(
+                    'collect_payment',
+                    plan_id=plan.id,
+                    subscriber=subscription.subscriber,
+                    amount=amount,
+                )
+                subscription.last_collected_time = current_period
+            else:
+                subscription.payment_failures += 1
+                if subscription.payment_failures > FailureTolerance:
+                    stop_subscription(plan.id, subscription.subscriber)
+                    emit Event(
+                        'stop_subscription',
+                        plan_id=plan.id,
+                        subscriber=subscription.subscriber,
+                    )
+    else:
+        break
+```
--- a/x/subscription/spec/README.md
+++ b/x/subscription/spec/README.md
@@ -1,0 +1,24 @@
+<!--
+order: 0
+title: Subscription Overview
+parent:
+  title: "subscription"
+-->
+
+# `x/subscription`
+
+## Table of Contents
+
+<!-- TOC -->
+
+1. **[Concepts](01_concepts.md)**
+2. **[State](02_state.md)**
+3. **[Messages](03_messages.md)**
+4. **[Events](04_events.md)**
+5. **[Params](05_params.md)**
+6. **[BeginBlock](06_begin_block.md)**
+
+## Abstract
+
+`x/subscription` is an implementation of a Cosmos SDK module that allows for billing customers for goods or services at
+regular intervals.
--- a/x/subscription/spec/04_events.md
+++ b/x/subscription/spec/04_events.md
@@ -1,0 +1,33 @@
+<!--
+order: 4
+-->
+
+# Events
+
+The `x/subscription` module emits the following events:
+
+## BeginBlocker
+
+| Type              | Attribute Key | Attribute Value |
+| ----------------- | ------------- | --------------- |
+| collect_payment   | subscriber    | {Address}       |
+| collect_payment   | amount        | {Coins}         |
+| collect_payment   | plan_id       | {int}           |
+| stop_subscription | plan_id       | {int}           |
+| stop_subscription | subscriber    | {Address}       |
+
+
+## MsgStopSubscription
+
+| Type              | Attribute Key | Attribute Value |
+| ----------------- | ------------- | --------------- |
+| stop_subscription | plan_id       | {int}           |
+| stop_subscription | subscriber    | {Address}       |
+
+## MsgCreateSubscription
+
+| Type                | Attribute Key | Attribute Value |
+| ------------------- | ------------- | --------------- |
+| create_subscription | plan_id       | {int}           |
+| create_subscription | subscriber    | {Address}       |
+
--- a/x/subscription/spec/05_params.md
+++ b/x/subscription/spec/05_params.md
@@ -1,0 +1,13 @@
+<!--
+order: 5
+-->
+
+# Parameters
+
+The subscription module contains the following parameters:
+
+| Key                 | Type | Example |
+| ------------------- | ---- | ------- |
+| GasPerCollection    | int  | "10000" |
+| SubscriptionEnabled | bool | true    |
+| FailureTolerance    | int  | 30      |


x/subscription/spec/01_concepts.md
```diff
@@ -0,0 +1,29 @@
...
+
+# Concepts
+
+## Subscription
+
+```go
+type Plan struct {
+  id int,  // unique identifier
+  owner Address,
+  price Coin,
```
tomtau, line 13: is it a fixed price or a max price?
for a floating exchange rate, one may want to be able to charge a different price each time

x/subscription/spec/README.md
```diff
@@ -0,0 +1,23 @@
...
+## Table of Contents
+
+<!-- TOC -->
+
+1. **[Concepts](01_concepts.md)**
+2. **[State](02_state.md)**
+3. **[Messages](03_messages.md)**
+4. **[Events](04_events.md)**
+5. **[Params](05_params.md)**
+6. **[BeginBlock](06_begin_block.md)**
```
tomtau, line 19: are these two to be added?

tomtau (changes requested):

x/subscription/spec/01_concepts.md
yihuang: merchant want to anchor to a fixed fiat currency price?

x/subscription/spec/01_concepts.md
```diff
@@ -0,0 +1,29 @@
...
+  interval int, // number of days
+}
+
+type Subscription struct {
```
tomtau, line 17: How is "expiration" going to be managed -- can merchant / payment gateway cancel/unsubscribe without cancelling the entire plan?

yihuang: merchant cancel some specific subscriber? it can also choose to stop the off-chain service if it can't collect payments from the subscriber, and resume the service as soon as payment resumes.

x/subscription/spec/01_concepts.md
tomtau: yep, that's a very common use case

yihuang: maybe introduce some oracle?

x/subscription/spec/01_concepts.md
tomtau: > merchant cancel some specific subscriber?
yes

> it can also choose to stop the off-chain service if it can't collect payments from the subscriber, and resume the service as soon as payment resumes.

that doesn't solve the problem, as it'll keep charging (the collection is per plan) / merchant (or payment gateway on its behalf) can't cancel individual subscriptions

x/subscription/spec/01_concepts.md
tomtau: merchants generally won't do the collection on their own if they receive fiat and it's via a payment provider -- the payment provider does the collection and sets the rate

yihuang: but if the payment provider is malicious, he can increase the price by change the rate without subscriber's agreement?

tomtau: generally, that's set in the subscription terms and conditions -- if that happens, the subscriber can cancel subscription and if it broke the terms and conditions, they can go to a small claims court etc.

x/subscription/spec/README.md
yihuang: removed

yihuang:
- merchant can maintain fiat-token conversion rate
- both merchant and subscriber can cancel subscription

x/subscription/spec/01_concepts.md
yihuang: both merchant and subscriber can cancel subscription now

x/subscription/spec/01_concepts.md
yihuang: done, payment provider or merchant can maintain the conversion rate.

x/subscription/spec/02_state.md
```diff
@@ -0,0 +1,34 @@
...
+```golang
+type Plan struct {
+  id int,  // auto-increasing unique identifier
+  title string,
+  description string,
+  owner Address,  // beneficial owner of the plan
+  price Decimal,  // price of fiat currency show to user
+  price_currency String,  // fiet currency name
+  rate Decimal,  // fiat-token conversion rate, maintained by owner
+
```
tomtau, line 18: rate to what ? what if e.g. one transfers in a stablecoin or a different asset over IBC and would want to create a subscription with that

tomtau (commented):
any network parameters? I guess it could have `subscriptionsenabled` similar to bank or ibc-transfer?

x/subscription/spec/02_state.md
yihuang: - added a `token_denomation` field.
- for stablecoin, merchant can maintain a near 1 conversion rate.

x/subscription/spec/02_state.md
```diff
@@ -0,0 +1,35 @@
...
+```golang
+type Plan struct {
+  id int,  // auto-increasing unique identifier
+  title string,
+  description string,
+  owner Address,  // beneficial owner of the plan
+  price Decimal,  // price in fiat currency show to user
+  fiat_currency String,  // fiat currency name
+  token_denom String,  // denomation of token
+  conversion_rate Decimal,  // conversion rate from fiat currency to token, maintained by owner
```
tomtau, line 18: probably good for @christopher-crypto to comment on this one -- maybe it'll be more practical to submit it each time with the collection message (so that there's less delay between rate setting and collection)?

x/subscription/spec/02_state.md
```diff
@@ -0,0 +1,35 @@
...
+
+```golang
+type Plan struct {
+  id int,  // auto-increasing unique identifier
+  title string,
+  description string,
+  owner Address,  // beneficial owner of the plan
+  price Decimal,  // price in fiat currency show to user
+  fiat_currency String,  // fiat currency name
+  token_denom String,  // denomation of token
```
tomtau, line 17: how is `String` different from `string`? good to make it clear with the respect to: https://github.com/crypto-com/chain-main/issues/336

I guess it can be made more general by being called e.g. `plan_denom` or `plan_currency` (instead of `fiat_currency`) and `subscriber_denom` (instead of `token_denom`) in the case there are use cases where the plan isn't denominated in fiat?

x/subscription/spec/02_state.md
```diff
@@ -0,0 +1,35 @@
...
+  conversion_rate Decimal,  // conversion rate from fiat currency to token, maintained by owner
+
+  interval int, // number of seconds of payment period
+  timezone Timezone, // the timezone used to offset the time interval
+  last_collection_period int, // period index = (timestamp + timezone_offset) / interval
```
tomtau, line 22: probably good for @christopher-crypto to comment on this one -- is ok to collect on the plan-level, or on the subscriber-level (given subscriptions start at different times)?

x/subscription/spec/02_state.md
```diff
@@ -0,0 +1,35 @@
...
+  title string,
+  description string,
+  owner Address,  // beneficial owner of the plan
+  price Decimal,  // price in fiat currency show to user
+  fiat_currency String,  // fiat currency name
+  token_denom String,  // denomation of token
+  conversion_rate Decimal,  // conversion rate from fiat currency to token, maintained by owner
+
+  interval int, // number of seconds of payment period
+  timezone Timezone, // the timezone used to offset the time interval
```
tomtau, line 21: timezones are definitely needed on the client / UX side -- for this network-only state, Unix time is enough (one can just use the interval and display it with the respect to the local settings) or will it create any problems?

yihuang: It's for calculating a period index for local timezone, since `block_time / interval` is different from `(block_time + 8hours) / interval`.

x/subscription/spec/02_state.md
```diff
@@ -0,0 +1,35 @@
...
+  last_collection_period int, // period index = (timestamp + timezone_offset) / interval
+}
```
christopher-crypto, line 23: Maybe we should have expiration_date as well - this would be for cases like phone plans where it's a subscription until a certain date

x/subscription/spec/02_state.md
```diff
@@ -0,0 +1,35 @@
...
+
+type Subscription struct {
+  planId int,
+  subscriber Address,
+}
```
christopher-crypto, line 28: Or maybe expiration_date should be added here

x/subscription/spec/03_messages.md
```diff
@@ -0,0 +1,94 @@
...
+Merchant or payment provider create plans.
+
+```protobuf
+message MsgCreatePlan {
+  string title;
+  string description;
+  Decimal price,  // price in fiat currency show to user
+  string fiat_currency,  // fiat currency name
+  string token_denom,  // denomination of token
+  Decimal conversion_rate,  // conversion rate from fiat currency to token, maintained by owner
```
tomtau, line 18: @christopher-crypto maybe good to have a possibility to specify 1-to-1 plans which will be automatically deducted at begin block (without the plan owner submitting collect payment msg)

tomtau: or simplify to only 1-to-1 rates (only specify denom, but no conversions)

tomtau: for "begin block" collection, advantage is no extra fees

x/subscription/spec/03_messages.md
```diff
@@ -0,0 +1,94 @@
...
+  int32 interval
```
tomtau, line 19: @thomasmktong-cryptocom maybe charging at "first day of the month" / "last day of the week" ...

tomtau: "first Monday of the month"

x/subscription/spec/02_state.md
tomtau: @christopher-crypto : real world may need both; but ok to start with the set interval

x/subscription/spec/03_messages.md
```diff
@@ -0,0 +1,94 @@
...
+message MsgStopSubscription {
+  int32 planId = 1,
+  Address subscriber = 2,
+}
+```
+
+The subscriber unsubscribe the plan, or plan owner cancel a subscription.
+
+## MsgCollectPayments
+
```
tomtau, line 68: @christopher-crypto in the real world, there's a "tolerance" -- where e.g. after 3 fails, it'll auto-cancel the subscription

tomtau: but both options exist -- e.g. merchant may not want to auto-cancel

x/subscription/spec/02_state.md
tomtau, line 22: maybe needs to be stored on the subscription level for a manual retry (if there's automated begin block collection)

x/subscription/spec/03_messages.md
```diff
@@ -0,0 +1,94 @@
...
+```protobuf
+message MsgCollectPayments {
+  int32 planId = 1,
+}
+```
+
+Can only issued by plan owner.
+
+`x/subscription` module automatically collect payments from subscribers at begin block, and write collection results into events. Merchants or payment providers can watch the events, if there's failed payment, they can choose to stop the off-chain service or cancel the subscription.
```
tomtau, line 77: for the automatic collection (without a manual tx by the plan owner), one thing to consider is how this could be misused -- as e.g. one could create plan-subscription cycles every second.
so one would either need to deduct fee for each operation or rate-limit the operations -- e.g. force that the plan has expiration time + there's a network param for "max collections"... so when creating the plan, one validation will be "(expiration time - time now) / period <= max collections".

one other random thought I had is that this could potentially be generalised not only for "msgsend", but for other tx that one may want to do periodically (withdraw reward, delegate...) -- but that's perhaps more advanced than the original goal / purpose of this module

yihuang: sounds like a standalone `cron` module ;D

tomtau: Weave (alternative to Cosmos SDK) has this module: https://github.com/iov-one/weave/tree/master/x/cron :D

tomtau: but that was just a random thought -- solving the original problem of payment subscriptions is useful on its own and generalising it may make it more complex than it needs to be

yihuang: Regarding the rate-limiting, I think we can consume some gas(https://docs.cosmos.network/master/basics/gas-fees.html) for CreatePlan message to cover the potential calculations later on.

x/subscription/spec/03_messages.md
yihuang: I think there's some complexity to add calendar to support non-fixed intervals, I guess we can support fixed-seconds intervals first, support calendar intervals at next step.

x/subscription/spec/03_messages.md
tomtau: the cost of the plan will depend on how many subscriptions there are -- so gas for creating subscription can take into account how many executions it'll lead to depending on the plan expiration and period?
given the expected initial gas price levels + exchange rate, not returning the "unused" gas from cancelled subscriptions shouldn't be a big issue for now -- similarly to what @christopher-crypto mentioned regarding the manual "retry" gas costs

yihuang: Added this:
```
period0 = (block_time + interval_offset) / interval
periodn = (expiration_time + interval_offset) / interval
ConsumeGas( (periodn - period0 + 1) * GasPerSubscription )
```

`GasPerSubscription` is a network parameter.

x/subscription/spec/02_state.md
yihuang: done

x/subscription/spec/01_concepts.md
yihuang: changed to fixed denomination.

x/subscription/spec/README.md
yihuang: added

x/subscription/spec/01_concepts.md
yihuang: both plan and subscription has expiration time now.

x/subscription/spec/02_state.md
yihuang: removed rate, use fixed token denomination.

x/subscription/spec/02_state.md
yihuang: Use `Coin` directly now, which should be well defined concept already I think.

x/subscription/spec/02_state.md
yihuang: stick to unified plan level collection time for now.

x/subscription/spec/02_state.md
yihuang: changed `timezone` to more general `interval_offset`, to specify intervals more flexible.

x/subscription/spec/02_state.md
yihuang: both plan and subscription support `expiration_time` now, automatically stopped when expired.

x/subscription/spec/02_state.md
yihuang: it's added in both plan and subscription now.

x/subscription/spec/03_messages.md
yihuang: now it supports:
- automatically collect in begin block once for each period
- merchant retry collection manually at any time

x/subscription/spec/03_messages.md
yihuang: Specified both intervals(fixed seconds and calendar one), but at implementation side, we can support second interval first for simplicity.

x/subscription/spec/03_messages.md
yihuang: Merchant can set `tolerate_failure` to `false`, then it'll automatically cancel failed subscription.
Or set `tolerate_failure` to `true`, then manually handle the retrying.

x/subscription/spec/02_state.md
yihuang: used fixed denomination now.

x/subscription/spec/01_concepts.md
```diff
@@ -0,0 +1,21 @@
+<!--
+order: 1
+-->
+
+# Concepts
+
+## Subscription
+
+- Merchant create/stop plans on chain, the price is defined on specified token denomination, users should do nessessary token conversion in other ways, e.g. exchange.
```
tomtau, line 9: just nit -- perhaps for the consistency, "plan owners" instead of merchants (e.g. this could potentially be used for regular donations).
```suggestion
- Subscription plan owners create/stop plans on Chain, the plan prices are defined for specified token denominations, subscribers should do necessary token conversions in other ways, e.g. exchange.
```

x/subscription/spec/01_concepts.md
```diff
@@ -0,0 +1,21 @@
...
+- User subscribe/unsubscribe to plan.
```
tomtau, line 10:
```suggestion
- Users subscribe/unsubscribe to plan.
```

x/subscription/spec/01_concepts.md
```diff
@@ -0,0 +1,21 @@
...
+- Payments are collected automatically at the start of specified interval, the results of collection are written into block events.
```
tomtau, line 11:
```suggestion
- Payments are collected automatically at the start of their specified intervals, the results of collections are written into block events.
```

x/subscription/spec/01_concepts.md
```diff
@@ -0,0 +1,21 @@
...
+- Create plan transaction will consume some gases to cover the computational cost of future automatic payment collections.
```
tomtau, line 12:
```suggestion
- Create plan transaction will consume a gas fee to cover the computational cost of future automatic payment collections.
```

x/subscription/spec/01_concepts.md
```diff
@@ -0,0 +1,21 @@
...
+- In case of automatic collection fails on some subscribers, the chain won't automatic retry later, merchant can either configure it to automatic cancel these subscribers, or retry collection later manually.
```
tomtau, line 13:
```suggestion
- If the automatic collection mechanism fails on some subscribers, Chain won't automatically retry later, the corresponding plan owner can either configure their plan to automatically cancel these subscribers or manually retry failed collections later.
```

x/subscription/spec/01_concepts.md
```diff
@@ -0,0 +1,21 @@
...
+
+### Intervals
+
+We support specify following types of intervals for plan:
+
+- Fixed seconds interval with offset, like "the beginning of each day offset by UTC+8".
+
+- [TODO] Calender interval, like "first day of each month".
```
tomtau, line 21: I'm thinking how this can work with future extensibility -- I guess instead of hardcoding integer, or perhaps have it in protobuf as:
```
message Interval {
  oneof value {
    FixedInterval               fixed                 = 1;
    CalendarInterval              calendar                = 2;
  }
}
```

but perhaps to leave it as it is now -- in the future when calendar intervals are added, one can just use the upgrade module and have the data migration that will convert the stored plans

x/subscription/spec/02_state.md
```diff
@@ -0,0 +1,44 @@
+<!--
```
tomtau, line 1: I guess ok for these to be Golang types or use protobuf for consistency?

x/subscription/spec/02_state.md
```diff
@@ -0,0 +1,44 @@
...
+The `x/subscription` stores the plans and subscriptions on chain:
+
+```golang
+type Plan struct {
+  id int,  // auto-increasing unique identifier
+  title string,
+  description string,
+  owner Address,  // beneficial owner of the plan
+  price Coin,  // price to pay for each period, Coin contains both amount and denomination
+  expiration_time int,  // the timestamp in seconds when plan expires, it will be removed automatically
```
tomtau, line 16: so this is like "global" expiration time and each subscription has its own expiration time? 
maybe both options are possible as per @christopher-crypto -- in here, maybe it'll make sense to have something like "duration" in Plan and subscription's expiration will be "block_time + plan duration" ?

x/subscription/spec/03_messages.md
```diff
@@ -0,0 +1,73 @@
+<!--
+order: 3
+-->
+
+# Messages
+
+## MsgCreatePlan
+
+Merchant create plans, the message signer is the plan owner.
```
tomtau, line 9:
```suggestion
Plan owners create plans, the message signer is the plan owner.
```

x/subscription/spec/03_messages.md
```diff
@@ -0,0 +1,73 @@
...
+  int32 expiration_time,
+  bool tolerate_failure,
+  int32 interval
+  int32 interval_offset
+}
+```
+
+## MsgStopPlan
+
+Sent by plan owner, also removes all subscriptions.
```
tomtau, line 25:
```suggestion
Sent by plan owners, also removes all corresponding subscriptions.
```

x/subscription/spec/03_messages.md
```diff
@@ -0,0 +1,73 @@
...
+```
+
+## MsgCreateSubscription
+
+The message signer subscribe to the plan, fails if it'll expire before next collection time.
+
+```protobuf
+message MsgCreateSubscription {
+  int32 plan_id,
+  int32 expiration_time,
```
tomtau, line 40: ok for subscribers to choose expiration time @christopher-crypto ?

x/subscription/spec/03_messages.md
```diff
@@ -0,0 +1,73 @@
...
+```protobuf
+message MsgStopSubscription {
+  int32 plan_id,
+  Address subscriber,
+}
+```
+
+## MsgReCollectPayments
+
+Plan owner can manually retry payment collection in case of failure. only those failed to pay in current period will be collected again, success ones are skipped.
```
tomtau, line 65:
```suggestion
Plan owner can manually retry payment collection in the case of failures. Only unpaid subscriptions in the current period will be collected again, paid ones are skipped.
```

x/subscription/spec/03_messages.md
```diff
@@ -0,0 +1,73 @@
...
+
+```
+period0 = (block_time + interval_offset) / interval
+periodn = (expiration_time + interval_offset) / interval
+ConsumeGas( (periodn - period0 + 1) * GasPerSubscription )
+```
+
+## MsgStopSubscription
+
+The subscriber or plan owner stop a subscription.
```
tomtau, line 54:
```suggestion
A subscriber or a plan owner can stop a subscription.
```

x/subscription/spec/06_begin_block.md
```diff
@@ -0,0 +1,56 @@
...
+    else:
+        break
+```
+
+## Collect payments
+
+`x/subscription` module automatically collect payments from subscribers at begin block, and write collection results into events.
+
+```python
+for plan_id, next_collectime_time in get_plans_sorted_by_next_collection_time():
```
tomtau, line 32: what is next collection time? perhaps iterate subscriptions directly?

tomtau (changes requested):
looking good! just left a few comments

x/subscription/spec/06_begin_block.md
yihuang: yeah, it seems makes more sense to iterate subscriptions directly, and in the new revision, failed subscription get retried automatically. I also makes the `FailureTolerance` to be a network parameter. failed subscription get stopped after retried some times.

tomtau: the cancellation option may still need to be there -- e.g. if the plan owner terminates the service for particular/selected subscribers (while still keeping the other subscriptions going), they may also want to proactively cancel their subscriptions regardless of their balance, so they don't get charged by mistake (as affected subscribers may forget to unsubscribe / cancel themselves)

tomtau (commented):
looks all right -- the only unsure thing for me is whether a subscriber should be able to choose expiration time, or there was a fixed duration specified in the plan? @christopher-crypto

x/subscription/spec/03_messages.md
```diff
@@ -0,0 +1,61 @@
...
+periodn = (expiration_time + interval_offset) / interval
+ConsumeGas( (periodn - period0 + 1) * GasPerCollection )
+```
+
+## MsgStopSubscription
+
+A subscriber or a plan owner can stop a subscription.
+
+```protobuf
+message MsgStopSubscription {
```
devashishdxt, line 57: If a user has a monthly subscription and they stop a subscription. Does the subscription gets immediately stopped or at the end of month (when they have already paid for the whole month)?

yihuang: It's immediately removed from the state, but if you query the events of success payment collection of current period index, it's still in that list, I think merchant can use that list as the valid subscription list.

x/subscription/spec/02_state.md
```diff
@@ -0,0 +1,52 @@
...
+  tolerate_failure bool, // if false, it'll automatically cancel subscriptions fails to pay, otherwise merchant can handle them manually
+
+  interval int, // payment interval defined in number of seconds
+  interval_offset int, // interval offset in terms of seconds to support timezones or any arbitrary time offsets
```
calvinaco, line 20: Can you share the use case of using `interval_offset` to help me better understanding this? Thanks

calvinlauco, line 20: Can you share a few usecases of `interval_offset` to help me better understand it. Thanks!

x/subscription/spec/02_state.md
```diff
@@ -0,0 +1,52 @@
...
+
+The `x/subscription` stores the plans and subscriptions on chain:
+
+```golang
+type Plan struct {
+  id int,  // auto-increasing unique identifier
+  title string,
+  description string,
+  owner Address,  // beneficial owner of the plan
+  price Coin,  // price to pay for each period, Coin contains both amount and denomination
```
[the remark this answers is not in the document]
tomtau: how would it work though? charge all of them or try charging one-by-one?

x/subscription/spec/02_state.md
yihuang: > Can you share a few usecases of `interval_offset` to help me better understand it. Thanks!

For example:
```
# 0-24 hour in UTC
period index = timestamp / (24 * 3600)
# 0-24 hour in UTC+8
period index = (timestamp + 8 * 3600) / (24 * 3600)

File: x/subscription/spec/02_state.md (hunk 31 above)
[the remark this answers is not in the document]
Comment (yihuang) [reply x/subscription/spec/02_state.md:15]: Charge a portfolio of coins?

File: x/subscription/spec/03_messages.md (hunk 29 above)
Comment (yihuang) [reply x/subscription/spec/03_messages.md:57]: We don't have a builtin way to query events directly though, unless using chain-indexing.

File: x/subscription/spec/03_messages.md [hunk 32]
```diff
@@ -0,0 +1,61 @@
+<!--
+order: 3
+-->
+
+# Messages
+
+## MsgCreatePlan
```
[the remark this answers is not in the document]
yihuang: Sounds good

x/subscription/spec/02_state.md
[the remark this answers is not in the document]
calvinlauco: It should charge all of them. It is consider as one single set of coins.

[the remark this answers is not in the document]
tomtau: not sure if there is any use case for that -- I'd favour the current simpler version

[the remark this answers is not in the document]
tomtau: @christopher-crypto may comment if it'd be useful to charge several currencies on a single subscription at a time?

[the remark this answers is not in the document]
calvinlauco: Yes, I think single Coin should work in most subscription use case, I also don't have sth in mind for mutliple coins at this moment.

The reason I suggest is I see other modules are doing it, so it is not hard to support (After all `Coins` is just an array of `Coin`, and the `Bank` keeper is supporting sending `Coins` instead of `Coin` actually (https://github.com/cosmos/cosmos-sdk/blob/master/x/bank/keeper/send.go#L19)

But I do agree because in fact on our current chain, all messages. e.g. `MsgSend`, we are always having only one token (`basetcro`) inside `Coins`.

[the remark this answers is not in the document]
yihuang: Since there's transfer for coins, then I think it's trivial to change it to `Coins`.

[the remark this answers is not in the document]
yihuang: done

x/subscription/spec/02_state.md
calvinlauco: Thanks

x/subscription/spec/01_concepts.md
yihuang: sorry I didn't noticed these comments, they are collapsed by default.
I'll see if I can manage to bring in some calendar support.

x/subscription/spec/02_state.md
yihuang: it seems cosmos has a pattern to use golang types in [state spec](https://github.com/cosmos/cosmos-sdk/blob/master/x/slashing/spec/02_state.md), but protobuf in [messages](https://github.com/cosmos/cosmos-sdk/blob/master/x/slashing/spec/03_messages.md).

x/subscription/spec/02_state.md
yihuang: this is the expiration time of the whole plan, the whole plan is stopped after this time.

x/subscription/spec/03_messages.md
yihuang: changed to use duration configured in plan.

x/subscription/spec/01_concepts.md
yihuang: Support crontab syntax directly, that provides best UX.

x/subscription/spec/02_state.md
yihuang: Plan don't expire now.

x/subscription/spec/02_state.md
yihuang: In new revision, it changed to use crontab syntax to specify intervals.

yihuang:
- added crontab syntax to specify intervals
- plan don't expire automatically
- subscriptions expire according to duration specified in plan

x/subscription/spec/02_state.md
```diff
@@ -0,0 +1,87 @@
...
+  price Coins,  // price to pay for each period, Coins contains both amount and denomination
+  subscription_duration int,  // duration of subscriptions
+  cron_spec CronSpec,  // Configure time intervals, parsed from crontab syntax
+  cron_tz Timezone,  // timezone for cron_spec
+}
+
+type Subscription struct {
+  plan_id int,
+  subscriber Address,
+  create_time int, // the block time when subscription created
```
tomtau, line 24:
```suggestion
  create_time int, // the block time when subscription was created
```

x/subscription/spec/02_state.md
```diff
@@ -0,0 +1,87 @@
...
+  // the timestamp of last successful collection,
+  // default to the current block time round-down against cron spec
+  // subscriber don't pay for the period it get created in
```
tomtau, line 27:
```suggestion
  // subscribers don't pay for the period it gets created in
```

tomtau (approved):
👍

PR #335, opened 2021-01-04
Problem: Empty genesis validation function

### LINKED ISSUE #333: Problem: empty genesis validation function
in genesis.go the `Validate()` function only returns "nil" -- the function should be updated to check the genesis state consistency or the comment should be added to explain why it doesn't perform any validations

devashishdxt:
Solution: Added a comment explaining the reason for not performing any validations. Fixes #333.


tomtau:
same here: https://github.com/crypto-com/chain-main/blob/master/x/chainmain/types/genesis.go ?

devashishdxt:
> same here: https://github.com/crypto-com/chain-main/blob/master/x/chainmain/types/genesis.go ?

Done

tomtau (approved):

PR #346, opened 2021-01-07
Problem: insufficient validation of signed integers in CLI arguments

### LINKED ISSUE #332: Problem: insufficient validation of signed integers in CLI arguments
primarily in genaccounts.go and testnet.go.
e.g. in vesting, the values are parsed are signed integers and checked for non-zero, but the arguments are never checked for a negative values prior to use. Thus, inappropriate values may be sent into downstream logic.

devashishdxt:
Solution: Changed types from signed to unsigned integers. Fixes #332.


x/chainmain/client/cli/genaccounts.go
```diff
@@ -82,24 +82,30 @@ contain valid denominations. Accounts may optionally be supplied with vesting pa
 			if errstart != nil {
 				return fmt.Errorf("failed to parse vesting start: %w", errstart)
 			}
-			vestingStart, errstart := strconv.ParseInt(vestingStartStr, 10, 64)
+			vestingStartUint, errstart := strconv.ParseUint(vestingStartStr, 10, 64)
```
tomtau, line 85: should the bitsize be 63 instead of 64 (as it's later casted to a signed int64)?

tomtau (commented):
not if this fix introduces potential overflows (when casting unsigned to signed numbers)

x/chainmain/client/cli/genaccounts.go
devashishdxt: Changed to 32 to avoid overflow. Because we are converting it into `int64` later.

tomtau (commented):
the integration test failed -- https://github.com/crypto-com/chain-main/pull/346/checks?check_run_id=1660486167#step:5:380 --- I'll try to rerun it with the latest master merged

tomtau (approved):

PR #349, opened 2021-01-08
Problem: codebase is not using the latest SDK pre-release

tomtau:
Solution: bumped the dependency to rc7 + removed the double amount
integration tests workaround


tomtau:
oops, will update the vendor sha256 in a moment

yihuang (approved):

PR #351, opened 2021-01-08
Problem: crossfire config not possible (fixes #350)

### LINKED ISSUE #350: Problem: crossfire config not possible
The crossfire network should use the "mainnet" parameters (HD paths, bech32 prefixes...), but "tcro" denomination to denote it's a test token

tomtau:
Solution: added another network build configuration option

### DIFF
--- a/config/prefix_crossfire.go
+++ b/config/prefix_crossfire.go
@@ -1,4 +1,4 @@
-// +build testnet,crossfire
+// +build !testnet,crossfire
 
 package config
 
--- a/config/prefix_crossfire.go
+++ b/config/prefix_crossfire.go
@@ -1,4 +1,4 @@
-// +build !testnet,!crossfire
+// +build testnet,crossfire
 
 package config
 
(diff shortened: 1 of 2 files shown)


tomtau:
just for the context, as mainly @calvinaco and @lezzokafka may be familiar with it -- crossfire will use the mainnet parameters (bech32 prefixes, HD path...), but should use the "test cro" token denomination

yihuang:
> just for the context, as mainly @calvinaco and @lezzokafka may be familiar with it -- crossfire will use the mainnet parameters (bech32 prefixes, HD path...), but should use the "test cro" token denomination

maybe it's more straightforward to make a new network type?

tomtau:
> > just for the context, as mainly @calvinaco and @lezzokafka may be familiar with it -- crossfire will use the mainnet parameters (bech32 prefixes, HD path...), but should use the "test cro" token denomination
> 
> maybe it's more straightforward to make a new network type?

yeah, perhaps -- I see the Cosmos SDK only has one "baseDenom", so currency conversions won't work for tcro

yihuang (approved):

config/prefix_crossfire.go
```diff
@@ -0,0 +1,20 @@
+// +build testnet,crossfire
```
devashishdxt, line 1: Shouldn't this be `!testnet`?

PR #353, opened 2021-01-11
Problem: relayer master is not used

yihuang:
Solution:
- Since related PRs are merged, we change to use master branch


tomtau (approved):
👍

PR #363, opened 2021-01-27
Problem: latest SDK is not used

tomtau:
Solution: bumped dependencies + renamed the upgrade handler
(as 0.41 has consensus breaking changes) for the testnet usage


yihuang (approved):

yihuang:
Test case `test_ibc` timeout when waiting for ibc channel setup, not sure if it's accidental, trying re-run.

devashishdxt (approved):

PR #370, opened 2021-02-04
Problem: upgrade for crossfire was not correct

tomtau:
Solution:
- updated the upgrade handler to what @lezzokafka put in the proposal
- updated SDK to a patched fork of v0.41 with the evidence / bech32 fix


yihuang (approved):

Gerana295:
tomtau:upgrade/fix-0.9

PR #377, opened 2021-02-16
Problem: memory leak in tendermint

tomtau:
Solution: upgraded the dependency to the latest 0.34.4 version


leejw51crypto (approved):
lgtm

PR #379, opened 2021-02-21
Problem: (fix #369) goreleaser tarballs aren't compatible with the current cosmovisor

### LINKED ISSUE #369: Problem: goreleaser tarballs aren't compatible with the current cosmovisor
As discussed / observed on Slack by @allthatjazzleo and @yihuang

allthatjazzleo:
Solution: add bin to wrap binary


tomtau (approved):
lgtm -- this should also fix the windows builds

PR #397, opened 2021-03-03
Problem: protobuf files are using v1beta from cosmos-sdk (fixes #395)

### LINKED ISSUE #395: Problem: protobuf files are using v1beta from cosmos-sdk
Version the protobuf files in proto directory properly.

Options([source](https://github.com/crypto-org-chain/chain-main/pull/381#discussion_r584483594)):
1. follow Cosmos SDK's current version: keep "v1beta1"?
2. follow chain-main's "global version" module -- mainnet base repo will be v1, so subscriptions will likely be "v2"?
3. each module versions independently -- so subscriptions can be "v1" regardless of what mainnet base repo is and will only be bumped to "v2" if there are breaking changes in subscriptions?

tomtau:
Solution:
- upgraded Cosmos SDK to a patched version of 0.40.4
- updated changelog
- updated the module to explicitely prefix with v1
- updated the custom protobufs to be v1 instead of v1beta1 and regenerated files using protocgen


app/app.go
```diff
@@ -82,13 +82,13 @@ import (
...
 	upgradekeeper "github.com/cosmos/cosmos-sdk/x/upgrade/keeper"
 	upgradetypes "github.com/cosmos/cosmos-sdk/x/upgrade/types"
-	appparams "github.com/crypto-org-chain/chain-main/app/params"
-	"github.com/crypto-org-chain/chain-main/x/chainmain"
-	chainmainkeeper "github.com/crypto-org-chain/chain-main/x/chainmain/keeper"
-	chainmaintypes "github.com/crypto-org-chain/chain-main/x/chainmain/types"
-	supply "github.com/crypto-org-chain/chain-main/x/supply"
-	supplykeeper "github.com/crypto-org-chain/chain-main/x/supply/keeper"
-	supplytypes "github.com/crypto-org-chain/chain-main/x/supply/types"
+	appparams "github.com/crypto-org-chain/chain-main/v1/app/params"
```
yihuang, line 85: What's the version here used for?

tomtau: that's to signal the global version / start the mainnet versioning -- afaik I guess for v1, it's optional, but we can start. See Cosmos Hub: https://github.com/cosmos/gaia/commit/2bb04266266586468271c4ab322367acbf41188f

calvinaco:
@tomtau Does it mean all the message type is renamed from `{module}/v1beta1/...` to `{module}/v1/...` now?

tomtau:
> @tomtau Does it mean all the message type is renamed from `{module}/v1beta1/...` to `{module}/v1/...` now?

not all -- the Cosmos SDK ones will still be v1beta1... this is only for our own types. In "v1", it only affects "supply" module, but that didn't have any messages (only GRPC endpoints)

calvinaco:
> > @tomtau Does it mean all the message type is renamed from `{module}/v1beta1/...` to `{module}/v1/...` now?
> 
> not all -- the Cosmos SDK ones will still be v1beta1... this is only for our own types. In "v1", it only affects "supply" module, but that didn't have any messages (only GRPC endpoints)

Got it, thanks for the clarification.

app/app.go
yihuang: why cosmos-sdk don't use this kind of versioning?

yihuang: Is it only for documenting purpose? I think it'll always build the version of code in current repo, no matter what the version number is?

leejw51crypto: just for protocol level, use versions.
in coding level, use just use it as configured as current
otherwise, it would be too complex in development.

leejw51crypto (approved):
lgtm

devashishdxt (approved):

calvinaco (approved):

linfeng-crypto (approved):

app/app.go
tomtau: @yihuang AFAIK Cosmos SDK is "pre-1.0", so they don't guarantee anything / they don't want to signal any long-term maintenance at this point. 

@leejw51crypto it's more for the future -- https://blog.golang.org/v2-go-modules

JayT106 (commented):
lgtm

ikeedge (commented):
.

ikeedge:
😀

ikeedge:
.

ALESYAkot (commented):
[****]()

PR #402, opened 2021-03-04
Problem: (fix #401)missing git submodule inside dockerfile

### LINKED ISSUE #401: Problem: missing git submodule inside dockerfile
if one build docker image without `git submode update` before docker command, it will fail due to lack of dependency.

allthatjazzleo:
Solution: add git module update command to dockerfile

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,6 +18,7 @@ COPY . .
 
 # Install minimum necessary dependencies, build Cosmos SDK, remove packages
 RUN apk add --no-cache $PACKAGES && \
+  git submodule update --init --recursive && \
   NETWORK=${NETWORK} make install
 
 # Final image


Dockerfile
```diff
@@ -18,6 +18,8 @@ COPY . .
 
 # Install minimum necessary dependencies, build Cosmos SDK, remove packages
 RUN apk add --no-cache $PACKAGES && \
+  git submodule init && \
+  git submodule update && \
```
leejw51crypto, line 22: this commands can be one
git submodule update --init

leejw51crypto (approved):
lgtm

Dockerfile
allthatjazzleo: fixed

tomtau (approved):

PR #403, opened 2021-03-05
Problem: no lightweight extensibility approach (fixes #392)

### LINKED ISSUE #392: Problem: no lightweight extensibility approach
As currently, TM doesn't support "soft forks", it may be a bit too disruptive to use the upgrade module too often for upcoming use cases (e.g. non-fungible digital assets)

tomtau:
Solution: added a proposal to include the CosmWasm module

### DIFF
--- a/doc/architecture/adr-001.md
+++ b/doc/architecture/adr-001.md
@@ -1,0 +1,124 @@
+# ADR 001: Add CosmWasm Module
+
+## Changelog
+* 05-03-2021: Initial Draft
+* 08-03-2021: Updated Backwards-compatibility Comments
+
+## Context
+
+Crypto.org Chain logic may need to be extended to support new digital asset and financial use cases
+in the near future. 
+In the initial implementation, Crypto.org Chain logic is extensible through three different mechanisms:
+
+1. Cryptographic commitments and multi-signatures: 
+through the existing and future (e.g. BIP-340) signature verification mechanisms, 
+one can potentially create accounts that, through their construction, enforce additional conditions
+on their transactions. This could done in different ways (e.g. see "Discreet Log Contracts" and "Adaptor Signatures").
+Alternatively, the most "flexible" approach is if a conventional multi-signature is constructed, but one of the
+"signers" is a secure enclave program executed by all parties.
+
+2. Inter-Blockchain Communication (IBC): the other connecting network or solo machine could enforce additional conditions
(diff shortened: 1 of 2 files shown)


doc/architecture/adr-001.md
```diff
@@ -0,0 +1,117 @@
...
+dominant approach of metered "stored procedures", popularized by Ethereum under the term "smart contracts". 
+Unlike on Ethereum, the target binary format (WebAssembly) is an official LLVM target 
+and mainstream programming languages (such as Rust) can be used for development.
+The overall design avoids many of the common pitfalls (e.g. reetrancy attacks) developers face on Ethereum.
+
+There are two main concerns regarding CosmWasm:
+
+1. It is a new unstable module: this is concerning from both the security perspective
+(e.g. DoS attack vectors in its execution engine) and the backwards compatibility perspective
+(stored bytecode that targeted the older module version may not execute correctly on its new version).
```
ethanfrey, line 42: There have been a number of breaking CosmWasm changes in the byte-code to host interface. After CosmWasm 0.11, we kept that interface stable for a few versions while doing major improvements to the tooling (0.11, 0.12, 0.13). All nodes must run the same version (eg. different gas costs) but the same wasm code would work on all of them.

We are preparing a 0.14 release with quite a few breaking changes that we have been waiting for for some time. 

After this, we aim to get feedback and maybe one more pass of API changes before stabilizing a 1.0 CosmWasm. Once we commit to CosmWasm 1.0, the wasm interface will be fixed and you will not have to worry about such breaking changes on upgrades.

doc/architecture/adr-001.md
```diff
@@ -0,0 +1,117 @@
...
+
+2. Cosmos SDK itself is not yet stable: before "1.0", there may breaking changes and "smart contracts"
```
ethanfrey, line 44: We designed the communication with the blockchain via an adapter layer. This worked quite well and we we able to run the same contracts unmodified in all versions from wasmd 0.11 (CosmWasm 0.11 + Cosmos SDK 0.39) to wasmd 0.15 (CosmWasm 0.13 + Cosmos SDK 0.41).

Providing a stable environment to contracts should not be dependent on Cosmos SDK 1.0

ethanfrey (commented):
Thanks for considering this. A few comments

yihuang (approved):

devashishdxt (approved):

PR #405, opened 2021-03-09
Problem: chain-main doesn't use the upstream Cosmos SDK release

tomtau:
Solution: updated the dependencies to use 0.42.0 instead of the patched 0.41.4

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "cosmos-sdk"]
 	path = third_party/cosmos-sdk
-	url = https://github.com/crypto-com/cosmos-sdk.git
+	url = https://github.com/cosmos/cosmos-sdk.git
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+*March 9, 2021*
+## v1.0.1
+A version based on the upstream release of Cosmos SDK 0.42.0.
+(Note that the SDK 0.42.0 release is nearly identical to the patched SDK fork 0.41.4 that was used in 1.0.0.)
+### Improvements
+* [399](https://github.com/crypto-com/chain-main/pull/399) atomic updates to genesis file
+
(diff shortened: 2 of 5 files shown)


devashishdxt (approved):

JayT106:
Question for the Cosmos SDK submodule setup. Initially, I thought the reason we put this submodule because the Crypto.com  Cosmos SDK repo has its custom modification and might keeping update the codes. If we are using the Cosmos SDK official release. Is the submodule still necessary?

devashishdxt:
> Question for the Cosmos SDK submodule setup. Initially, I thought the reason we put this submodule because the Crypto.com Cosmos SDK repo has its custom modification and might keep updating the code. If we are using the Cosmos SDK official release. Is the submodule still necessary?

Which submodule are you talking about?

JayT106:
> > Question for the Cosmos SDK submodule setup. Initially, I thought the reason we put this submodule because the Crypto.com Cosmos SDK repo has its custom modification and might keep updating the code. If we are using the Cosmos SDK official release. Is the submodule still necessary?
> 
> Which submodule are you talking about?

the submodule "cosmos-sdk"

devashishdxt:
> the submodule "cosmos-sdk"

Ah. You mean git submodule?

JayT106:
> > the submodule "cosmos-sdk"
> 
> Ah. You mean git submodule?

Yes. However, just checked the project structure. Looks like `Nix` and `gPRC` need the submodule.

JayT106 (approved):
LGTM

tomtau:
> Question for the Cosmos SDK submodule setup. Initially, I thought the reason we put this submodule because the Crypto.com Cosmos SDK repo has its custom modification and might keep updating the code. If we are using the Cosmos SDK official release. Is the submodule still necessary?

I think it's still necessary for the protobuf generation -- e.g. building the integration tests needs to build gRPC clients

yihuang (approved):

Issue #398, opened 2021-03-03
Problem: genesis file is updated non-atomically


tomtau:
If the machine experiences a sudden fault, such as a power outage, while running the `init` command, the genesis file will be corrupted.
The atomic update be accomplished by writing the file to a temporary output (such as `genesis.json.tmp`), syncing that output using `file.Sync()`, calling `os.Rename` to rename the `genesis.json.tmp` to `genesis.json`, and finally applying a final `Sync()`.
https://github.com/crypto-org-chain/chain-main/blob/master/cmd/chain-maind/app/app.go#L136

yihuang:
Another approach is to patch it in `NewDefaultGenesisState`(`app/genesis.go`), then no need to override the command.

PR #409, opened 2021-03-10
Problem: chain-main can't be built for windows (fix #406)

### LINKED ISSUE #406: Problem: chain-main can't be built for windows
```
 failed after 216.97s error=failed to build for windows_amd64: # github.com/crypto-org-chain/chain-main/v1/cmd/chain-maind/app
111
cmd/chain-maind/app/app.go:160:10: undefined: renameio.WriteFile
```

leejw51crypto:
Solution: use ioutil.WriteFile,
file atomicity is done via file level.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,7 @@ jobs:
     
     strategy:
       matrix:
-        os: [macos-latest, ubuntu-latest, windows-latest]
+        os: [ubuntu-latest, windows-latest]
 
     timeout-minutes: 5
     if: >-
@@ -80,11 +80,6 @@ jobs:
       - name: build
         if: | 
           startsWith(matrix.os, 'ubuntu') &&
-          (env.GIT_DIFF != '' || github.event_name == 'issue_comment')
-        run: make build
-      - name: buildmac
-        if: |
-          startsWith(matrix.os, 'mac') &&
           (env.GIT_DIFF != '' || github.event_name == 'issue_comment')
         run: make build
       - name: buildwindows
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,7 +12,12 @@ on:
 jobs:
   build:
     name: Build
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    
+    strategy:
+      matrix:
+        os: [macos-latest, ubuntu-latest, windows-latest]
+
     timeout-minutes: 5
     if: >-
       github.event_name == 'push' || github.event_name == 'pull_request' || 
@@ -73,8 +78,20 @@ jobs:
         run: |
           echo "$(go env GOPATH)/bin" >> $GITHUB_PATH
       - name: build
+        if: | 
+          startsWith(matrix.os, 'ubuntu') &&
+          (env.GIT_DIFF != '' || github.event_name == 'issue_comment')
         run: make build
-        if: "env.GIT_DIFF != '' || github.event_name == 'issue_comment'"
+      - name: buildmac
+        if: |
+          startsWith(matrix.os, 'mac') &&
+          (env.GIT_DIFF != '' || github.event_name == 'issue_comment')
+        run: make build
+      - name: buildwindows
+        if: |
+          startsWith(matrix.os, 'windows') &&
+          (env.GIT_DIFF != '' || github.event_name == 'issue_comment')
+        run: make buildwindows  
       - name: test
         run: make test
         if: "env.GIT_DIFF != '' || github.event_name == 'issue_comment'"
--- a/Makefile
+++ b/Makefile
@@ -45,6 +45,11 @@ install: check-network go.sum
 
 build: check-network go.sum
 		go build -mod=readonly $(BUILD_FLAGS) $(BUILD_TAGS) -o $(BUILDDIR)/chain-maind ./cmd/chain-maind
+
+buildwindows: check-network go.sum
+		go build -buildmode=exe -mod=readonly $(BUILD_FLAGS) $(BUILD_TAGS) -o $(BUILDDIR)/chain-maind ./cmd/chain-maind
+
+
 .PHONY: build
 
 build-linux: go.sum
--- a/cmd/chain-maind/app/app.go
+++ b/cmd/chain-maind/app/app.go
@@ -38,7 +38,6 @@ import (
 	"github.com/crypto-org-chain/chain-main/v1/app/params"
 	"github.com/crypto-org-chain/chain-main/v1/config"
 	chainmaincli "github.com/crypto-org-chain/chain-main/v1/x/chainmain/client/cli"
-	"github.com/google/renameio"
 )
 
 // NewRootCmd creates a new root command for chain-maind. It is called once in the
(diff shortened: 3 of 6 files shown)


cmd/chain-maind/app/app.go
```diff
@@ -38,7 +39,6 @@ import (
 	"github.com/crypto-org-chain/chain-main/v1/app/params"
 	"github.com/crypto-org-chain/chain-main/v1/config"
 	chainmaincli "github.com/crypto-org-chain/chain-main/v1/x/chainmain/client/cli"
-	"github.com/google/renameio"
```
tomtau, line 41: maybe also remove the dependency from go.mod?

leejw51crypto: ok, i'll check

tomtau:
vendorSha256 may need to be updated for integration tests: https://github.com/crypto-org-chain/chain-main/pull/409/checks?check_run_id=2074911879 ?

JayT106:
So the root cause was the `renameio` implementation not compatible with the windows system.
And the CI doesn't have the binary build check across platforms?

tomtau:
> So the root cause was the `renameio` implementation not compatible with the windows system.
> And the CI doesn't have the binary build check across platforms?

That's correct -- CI only had Windows build for binary releases, so that's where this got caught: https://github.com/crypto-org-chain/chain-main/runs/2063584065?check_suite_focus=true#step:4:111

I think for this PR, it'll also be good to extend the existing CI workflow ( https://github.com/crypto-org-chain/chain-main/blob/master/.github/workflows/build.yml ) or add a new one (as that "build" workflow has many other steps that one doesn't need to worry about for simple "does it build on anything else than Linux?" check)

It should be possible with
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on

tomtau:
conflict with the latest master

leejw51crypto:
rebased

leejw51crypto:
add windows, mac build step

tomtau:
it seems the Windows "build" doesn't get executed: https://github.com/crypto-org-chain/chain-main/pull/409/checks?check_run_id=2084141603#step:9:3 it'll just checkout the code, but won't try to build it

tomtau:
@allthatjazzleo

leejw51crypto:
i'll check windows, mac build

tomtau:
builds now, but fails with:
```
# github.com/crypto-org-chain/chain-main/v1/cmd/chain-maind
97
C:\hostedtoolcache\windows\go\1.15.5\x64\pkg\tool\windows_amd64\link.exe: running gcc failed: exit status 1
98
C:/ProgramData/Chocolatey/lib/mingw/tools/install/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/8.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: Error: export ordinal too large: 69810
99
collect2.exe: error: ld returned 1 exit status
100
```

leejw51crypto:
that's golang issue for windows

applied this flag `buildmode`
 build  -buildmode=exe -mod=readonly $(BUILD_FLAGS) $(BUILD_TAGS) -o $(BUILDDIR)/chain-maind ./cmd/chain-maind

Makefile
```diff
@@ -45,6 +45,11 @@ install: check-network go.sum
 
 build: check-network go.sum
 		go build -mod=readonly $(BUILD_FLAGS) $(BUILD_TAGS) -o $(BUILDDIR)/chain-maind ./cmd/chain-maind
+
+buildwindows: check-network go.sum
+		go build  -buildmode=exe -mod=readonly $(BUILD_FLAGS) $(BUILD_TAGS) -o $(BUILDDIR)/chain-maind ./cmd/chain-maind
```
allthatjazzleo, line 50: an extra space there after `go build`

.github/workflows/build.yml
```diff
@@ -73,8 +78,14 @@ jobs:
         run: |
           echo "$(go env GOPATH)/bin" >> $GITHUB_PATH
       - name: build
+        if: startsWith(matrix.os, 'ubuntu')
         run: make build
-        if: "env.GIT_DIFF != '' || github.event_name == 'issue_comment'"
```
tomtau, line 77: if this condition isn't there, it may break the "/runsim" simulation workflow -- @allthatjazzleo may comment?

perhaps for windows/Mac -- instead of changing "build.yml", create a separate workflow ?

allthatjazzleo: This line shall not break "/runsim", it was to make sure the workflow will run only when there are some changes in `.go .mod .sum` file.
But since `technote-space/get-diff-action` step will empty result when event is `issue_comment`. I needed to put the second condition for event triggered by `/runsim` manually.

cmd/chain-maind/app/app.go
```diff
@@ -157,7 +157,7 @@ func initRootCmd(rootCmd *cobra.Command, encodingConfig params.EncodingConfig) {
 			return err
 		}
 
-		return renameio.WriteFile(cleanedPath, bz, 0600)
+		return ioutil.WriteFile(cleanedPath, bz, 0600)
```
yihuang, line 160: We need to reopen this issue then. https://github.com/crypto-org-chain/chain-main/issues/398

tomtau: seems not possible on Windows: https://github.com/google/renameio#windows-support ?

tomtau: or keep using renameio, just pin the dependency to be before https://github.com/google/renameio/commit/8b2b09a2cfe0f9be5e3b41bdf85c0baf422c0f6f knowing it's not reliable on Windows?

yihuang: is it possible to conditional compile?

tomtau: it should be: https://golang.org/cmd/go/#hdr-Build_constraints

so I guess one other option is to separate it out and `// +build !windows` on "app.go"-subset and have some Windows "app.go"-subset ?

tomtau: @leejw51crypto

Makefile
leejw51crypto: ok, i'll modify

cmd/chain-maind/app/app.go
leejw51crypto: ok, i'll apply conditional compile

leejw51crypto:
working now

leejw51crypto:
renameio for darwin, linux
ioutil for window

rebased

go.mod
```diff
@@ -8,7 +8,7 @@ require (
 	github.com/cosmos/ledger-go v0.9.2 // indirect
 	github.com/gogo/protobuf v1.3.3
 	github.com/golang/protobuf v1.4.3
-	github.com/google/renameio v1.0.0
+	github.com/google/renameio v0.1.0
```
tomtau, line 11: keep 1.0.0?

tomtau (changes requested):
I guess two options:
1. have just one "app.go" and use the old renameio
2. have two "app.go" and use the latest renameio -- just add build tags on the subset

for the second one, it'll be good to extract out the common functionality

go.mod
leejw51crypto: ok, i will go for option2, change version , too.

.github/workflows/build.yml
leejw51crypto: restored back , if condtions

tomtau (approved):

Issue #385, opened 2021-02-24
Problem: unstable join validator integration test


yihuang:
https://github.com/crypto-org-chain/chain-main/pull/381/checks?check_run_id=1968276739#step:5:349

yihuang:
https://github.com/crypto-org-chain/chain-main/pull/387/checks?check_run_id=1969411637#step:5:440

tomtau:
> https://github.com/crypto-org-chain/chain-main/pull/387/checks?check_run_id=1969411637#step:5:440

that one seems to be https://github.com/crypto-org-chain/chain-main/issues/388 -- is this a duplicate issue?

yihuang:
> > https://github.com/crypto-org-chain/chain-main/pull/387/checks?check_run_id=1969411637#step:5:440
> 
> that one seems to be #388 -- is this a duplicate issue?

yes

tomtau:
Here IBC test failed: https://github.com/crypto-org-chain/chain-main/pull/390/checks?check_run_id=1976907366#step:5:8133
again the same?

yihuang:
> Here IBC test failed: https://github.com/crypto-org-chain/chain-main/pull/390/checks?check_run_id=1976907366#step:5:8133
> again the same?

I think so, something running very slow I guess.

huangxufu:
I guess we can close this one now, seems pretty stable recently.

yihuang:
I think this one can be closed now, pretty stable recently.

PR #410, opened 2021-03-10
Problem: no Byzantine integration test(fix #365) 

### LINKED ISSUE #365: Problem: no Byzantine integration test
I guess could be similar to https://github.com/crypto-com/thaler/blob/main/integration-tests/multinode/byzantine_test.py
after the bech32 fix is merged

linfeng-crypto:
- create 3 validator nodes, keep node0 has more than  2/3 voting power
- stop node2, and copy node1's validator key file to node2
- start node2, waiting for node1/node2 jailed
- check the tokens after node slashed and byzantine error

### DIFF
--- a/integration_tests/configs/byzantine.yaml
+++ b/integration_tests/configs/byzantine.yaml
@@ -2,10 +2,8 @@ chain_id_1:
   validators:
     - coins: 8000cro
       staked: 8000cro
-      base_port: 26650
     - coins: 2000cro
       staked: 2000cro
-      mnemonic: wage frame tank harvest item cool oven other excite expand permit rescue card together absurd vapor water pole drip tomorrow clog illegal off column
     - coins: 1000cro
       staked: 1000cro
   accounts:
--- a/integration_tests/configs/byzantine.yaml
+++ b/integration_tests/configs/byzantine.yaml
@@ -1,0 +1,26 @@
+chain_id_1:
+  validators:
+    - coins: 8000cro
+      staked: 8000cro
+      base_port: 26650
+    - coins: 2000cro
+      staked: 2000cro
+      mnemonic: wage frame tank harvest item cool oven other excite expand permit rescue card together absurd vapor water pole drip tomorrow clog illegal off column
+    - coins: 1000cro
+      staked: 1000cro
+  accounts:
+    - name: wallet-1
+      coins: 20000cro
+  genesis:
+    app_state:
+      staking:
+        params:
+          unbonding_time: "10s"
+      gov:
+        voting_params:
+          voting_period: "10s"
+        deposit_params:
+          max_deposit_period: "10s"
+          min_deposit:
+            - denom: "basecro"
+              amount: "10000000"
--- a/integration_tests/test_byzantine.py
+++ b/integration_tests/test_byzantine.py
@@ -32,6 +32,10 @@ def test_byzantine(cluster):
     assert len(cluster.validators()) == 3
     from_node = 1
     to_node = 2
+    val_addr_byzantine = cluster.address("validator", i=from_node, bech="val")
+    tokens_byzantine_before = int((cluster.validator(val_addr_byzantine))["tokens"])
+    val_addr_slash = cluster.address("validator", i=to_node, bech="val")
+    tokens_slash_before = int((cluster.validator(val_addr_slash))["tokens"])
     cluster.stop_node(to_node)
     cluster.copy_validator_key(from_node, to_node)
     cluster.start_node(to_node)
@@ -65,3 +69,8 @@ def test_byzantine(cluster):
     assert val_2["jailed"]
     assert val_2["status"] == "BOND_STATUS_UNBONDING"
     print("\n{}s waiting for node 2 jailed".format(i))
+
+    tokens_byzantine_after = int((cluster.validator(val_addr_byzantine))["tokens"])
+    tokens_slash_after = int((cluster.validator(val_addr_slash))["tokens"])
+    assert tokens_byzantine_before * 0.95 == tokens_byzantine_after
+    assert tokens_slash_before * 0.99 == tokens_slash_after
--- a/integration_tests/test_byzantine.py
+++ b/integration_tests/test_byzantine.py
@@ -1,0 +1,67 @@
+import sys
+import time
+from pathlib import Path
+
+import pytest
+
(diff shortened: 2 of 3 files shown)


lezzokafka:
Should we also include the slashing part in this integration test?

tomtau (changes requested):
> Should we also include the slashing part in this integration test?

yeah, will be good to add test assertions that the account balance was decreased / slashed

tomtau:
I guess the integration test fail isn't relevant to this PR: https://github.com/crypto-org-chain/chain-main/pull/410/checks?check_run_id=2082753021#step:5:2026 ? is it https://github.com/crypto-org-chain/chain-main/issues/385 @yihuang  ?

yihuang (approved):
LGTM even better if we can assert the reason of jailing

yihuang:
> I guess the integration test fail isn't relevant to this PR: https://github.com/crypto-org-chain/chain-main/pull/410/checks?check_run_id=2082753021#step:5:2026 ? is it #385 @yihuang ?

yeah, i think so.

tomtau:
there are two small lint issues in the Python script, but otherwise good to be merged

devashishdxt (approved):

tomtau:
@yihuang it seems the IBC test consistently fails on this PR -- due to the address being in use: https://github.com/crypto-org-chain/chain-main/pull/410/checks?check_run_id=2083323245#step:5:997

is it that the node doesn't get stopped here?

linfeng-crypto:
> LGTM even better if we can assert the reason of jailing

I tried, but can't the details jailed reason, it seems that there is no API to get the reason.

integration_tests/test_byzantine.py
```diff
@@ -0,0 +1,76 @@
...
+
+    # it may take 30s to finish the loop
+    i = 0
+    while i < 300:
+        time.sleep(1)
+        sys.stdout.write(".")
+        sys.stdout.flush()
+        i += 1
+        val_addr = cluster.address("validator", from_node, bech="val")
+        val_1 = cluster.validator(val_addr)
```
leejw51crypto, line 51: how about val_1 -> validator1

integration_tests/test_byzantine.py
```diff
@@ -0,0 +1,76 @@
...
+        if val_1["jailed"]:
+            break
+    assert val_1["jailed"]
+    assert val_1["status"] == "BOND_STATUS_UNBONDING"
+    print("\n{}s waiting for node 1 jailed".format(i))
+
+    i = 0
+    # it may take 2min to finish the loop
+    while i < 300:
```
leejw51crypto, line 60: declare as const , in move to the top of source code?

leejw51crypto (approved):
lgtm,
small naming suggestions

JayT106:
LGTM, just thinking could we just use one while loop to check the status on both nodes? Does the consensus behavior always jail the node1 first and then jail the node2?

tomtau (changes requested):
it seems "test_ibc" consistently fails on this PR: https://github.com/crypto-org-chain/chain-main/pull/410/checks?check_run_id=2086193872#step:5:2085
```
ibc-0/node0.log 12:42PM INF Starting Node service impl=Node
1509
ibc-0/node0.log Error: failed to listen on 0.0.0.0:26657: listen tcp 0.0.0.0:26657: bind: address already in use
1510
ibc-0/node0.log Usage:
1511
ibc-0/node0.log   chain-maind start [flags]
1512
ibc-0/node0.log 
...
```

so either disable/ignore that test for now or fix it -- @yihuang may help?

yihuang:
> it seems "test_ibc" consistently fails on this PR: https://github.com/crypto-org-chain/chain-main/pull/410/checks?check_run_id=2086193872#step:5:2085
> 
> ```
> ibc-0/node0.log 12:42PM INF Starting Node service impl=Node
> 1509
> ibc-0/node0.log Error: failed to listen on 0.0.0.0:26657: listen tcp 0.0.0.0:26657: bind: address already in use
> 1510
> ibc-0/node0.log Usage:
> 1511
> ibc-0/node0.log   chain-maind start [flags]
> 1512
> ibc-0/node0.log 
> ...
> ```
> 
> so either disable/ignore that test for now or fix it -- @yihuang may help?

relayer execute failed:
```
E               subprocess.CalledProcessError: Command '['relayer', '--home', PosixPath('/tmp/pytest-of-runner/pytest-0/popen-gw0/data1/relayer'), 'light', 'init', 'ibc-0', '-f']' returned non-zero exit status 1.
```

not sure why.

pystarport/pystarport/cluster.py
```diff
@@ -698,7 +740,7 @@ def init_cluster(
                 config.keys(), list(rly_section.get("paths", {}).keys())
             ),
         )
-    if len(chains) > 1:
+    if config.get("relayer"):
```
yihuang, line 743: this change maybe the closest to the relayer issue in integration tests, can try to revert it and try again.

integration_tests/test_byzantine.py
linfeng-crypto: Ok

integration_tests/test_byzantine.py
linfeng-crypto: Ok

pystarport/pystarport/cluster.py
linfeng-crypto: Ok

tomtau:
still https://github.com/crypto-org-chain/chain-main/runs/2092934129#step:5:2049 🥺

integration_tests/configs/byzantine.yaml
```diff
@@ -0,0 +1,26 @@
+chain_id_1:
+  validators:
+    - coins: 8000cro
+      staked: 8000cro
+      base_port: 26650
```
yihuang, line 5: I guess this is the issue, don't set port explicitly, it'll be generated automatically.

tomtau: @linfeng-crypto

integration_tests/test_j_byzantine.py
```diff
@@ -0,0 +1,78 @@
+import sys
```
yihuang, line 1: is there a typo in the filename?

tomtau (approved):

PR #411, opened 2021-03-11
Problem: bank keeper isn't correctly initialized

tomtau:
Solution: updated the constructor
as per https://github.com/cosmos/cosmos-sdk/commit/3de375248759d57fb568b90d1bdb9a5ff9ef8a52


yihuang (approved):
I guess the issue is `distrtypes.ModuleName` was in the whitelist before, but now all module accounts are disallowed?

PR #414, opened 2021-03-11
Problem: goreleaser config not releasing for Arm64 (fix #376)

### LINKED ISSUE #376: Problem: goreleaser config not releasing for Arm64
Darwin arm64 for M1
Linux arm64 for AWS Graviton etc.

leejw51crypto:
Solution: upgrade golang to 1.16.1
to compile for apple m1


tomtau (changes requested):
the goreleaser here isn't still configured to release arm64? or would you want to split it into 2 PRs (upgrade go toolchain and update goreleaser settings)?

leejw51crypto:
yes, i'll add build step for apple m1.

leejw51crypto:
## osxcross links 
**osxcross/target/bin/**
1. **ls -la arm64-apple-darwin20.2-clang**
arm64-apple-darwin20.2-clang -> x86_64-apple-darwin20.2-wrapper
2. **ls -la x86_64-apple-darwin20.2-clang**
x86_64-apple-darwin20.2-clang -> x86_64-apple-darwin20.2-wrapper
3. **ls -la o64-clang**
o64-clang -> x86_64-apple-darwin20.2-wrapper

.goreleasermac.yml
```diff
@@ -0,0 +1,29 @@
+before:
```
tomtau, line 1: is it necessary to have this duplicate config? why not just ".goreleaser.yml"

.github/workflows/goreleasemac.yml
```diff
@@ -0,0 +1,31 @@
+name: goreleasermac
```
tomtau, line 1: why this separate pipeline? can't one just extend "make release-dry-run" and "make release" used in https://github.com/crypto-org-chain/chain-main/blob/master/.github/workflows/goreleaser.yml ?

tomtau (changes requested):
what about Linux arm64?

.github/workflows/goreleasemac.yml
leejw51crypto: troian/golang-cross doesn't support arm64+ darwin, yet
```
    goos:
      - darwin
    goarch:
      - arm64
      
```

leejw51crypto: error is  like this
image

tomtau: It supports Linux-arm64 though -- so perhaps, the easiest solution may be just to extend the current goreleaser config for Linux-arm64 (so one can deploy e.g. on AWS Graviton https://pages.awscloud.com/rs/112-TZM-766/images/2020_0501-CMP_Slide-Deck.pdf ) ... for M1, there are no inexpensive "servers" one can easily deploy to, so not having official binaries isn't a big problem -- for development, one can build them themselves or use Rosetta.
Later, when the "Darwin-arm64" is supported on troian/golang-cross, we can just extend the config

leejw51crypto: adding to current yml is possible, i'll include it.

.goreleasermac.yml
leejw51crypto: i'll remove .gorelasemac.yml, move it to current yml.

leejw51crypto: linux arm64 is possible, i'll add it.

.github/workflows/goreleasemac.yml
leejw51crypto: yes, i'll add linux arm64, that would be done very easily.
apple m1 binary is tested successfully, i'll submit to this pr

leejw51crypto:
linux arm64 not working correctly, fixing now

leejw51crypto:
linux arm64( arm8), mac m1 all working
 
for linux arm64
      - CGO_ENABLED=1
      - CC=x86_64-w64-mingw32-gcc
      - CXX=x86_64-w64-mingw32-g++

tomtau (approved):

leejw51crypto:
image

leejw51crypto:
temporarily applied #406, and built in other repo

Issue #422, opened 2021-03-22
Problem: multi-sign batch command is not exposed and tested


tomtau:
the command is in Cosmos SDK under "auth.GetMultiSignBatchCmd" -- it can read signatures from files and generate the multisig tx

PR #438, opened 2021-03-30
Problem: multi-sign batch command is not exposed and tested (#422)

linfeng-crypto:
solution: add `multisign-batch` in tx subcommand and integration test


tomtau (approved):
lgtm

tomtau (changes requested):
the liquid supply test fails: https://github.com/crypto-org-chain/chain-main/pull/438/checks?check_run_id=2224687315#step:5:389
I guess because 4000cro were added to the default network?

linfeng-crypto:
> the liquid supply test fails: https://github.com/crypto-org-chain/chain-main/pull/438/checks?check_run_id=2224687315#step:5:389
> I guess because 4000cro were added to the default network?

I'll check

tomtau (approved):

PR #441, opened 2021-03-31
Problem: Adding chain-maind keys will not be actually stored in disk …

leejw51crypto:
Solution: add decoding _hexcode in service name fetched


leejw51crypto:
keyring pr is here
https://github.com/99designs/keyring/pull/83/files

JayT106:
I see some of the modules been removed from the `go.sum`. How do they relate to the fixes?
For the keyring dependency, is it better to merge @leejw51crypto 's fixes to the upstream repo first and then use it? (Not sure how urgent is the fixes)

tomtau:
yeah, may be better to wait for the upstream -- it seems it's maintained

leejw51crypto:
> I see some of the modules been removed from the `go.sum`. How do they relate to the fixes?
> For the keyring dependency, is it better to merge @leejw51crypto 's fixes to the upstream repo first and then use it? (Not sure how urgent is the fixes)

i ran "go mod tidy" to check mod

this is urgent issue, because in linux,  keyring is not usable by default.

tomtau (changes requested):
there are some conflicts with the latest master.

it seems upstream may take time -- I've forked the repo with the fix here, so the go.mod can be updated to this one: https://github.com/crypto-org-chain/keyring/tree/gnome for the moment

leejw51crypto:
ok, i'll rebase with the latest

go.mod
```diff
@@ -29,3 +29,5 @@ replace google.golang.org/grpc => google.golang.org/grpc v1.33.2
 replace github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
 
 replace github.com/cosmos/ledger-cosmos-go => github.com/crypto-com/ledger-cosmos-go v0.9.10-0.20200929055312-01e1d341de0f
+
+replace github.com/99designs/keyring => github.com/leejw51crypto/keyring v1.1.6-gnome
```
tomtau, line 33:
```suggestion
replace github.com/99designs/keyring => github.com/crypto-org-chain/keyring v1.1.6-gnome
```

leejw51crypto:
inspecting integration test failures

leejw51crypto:
fixed, increase waiting time twice :-)

tomtau (approved):

matias904 (changes requested):
1150696e9218c9c6b615b3fdd0de1d1ccc8c1fbb

PR #449, opened 2021-04-06
Problem: error ambiguity in BitSet.Test(fix #428)

### LINKED ISSUE #428: Problem: error ambiguity in BitSet.Test
> This function conflates errors and legitimate bit testing results. Once the function returns false, there’s no way for the caller to know whether the passed index was out of bounds or the bit was not set. It is assumed that the development process took a shortcut in this case,since currently the `BitSet.Test` function is only invoked in `CompiledCronSpec.Roundup` and the parameter cannot be greater than 64. It would make sense to rewrite this function to avoid any potential mistakes in future code iterations.
> Recommendation: Rewrite the function to return a boolean and an error. If the supplied argument is out ofbounds, set the error.

linfeng-crypto:
solution: Rewrite the `BitSet.Test` function to return a boolean and an error


x/subscription/types/bitset.go
```diff
@@ -69,11 +69,11 @@ func (b *BitSet) Clear(i uint) error {
 
 // Test whether bit i is set.
 // if i >= 64, return false
```
yihuang, line 71: The comment need to change too then.

x/subscription/types/cronspec.go
```diff
@@ -407,14 +408,14 @@ func roundUpItem(spec CompiledCronItem, v uint32) uint32 {
 
 // CountPeriods counts how many datetimes between beginTime expirationTime matches the cron spec.
 // beginTime is not inclusive.
-func (spec CompiledCronSpec) CountPeriods(beginTime, expirationTime int64, tzoffset int32) (count uint64) {
-	count = 0
+func (spec CompiledCronSpec) CountPeriods(beginTime, expirationTime int64, tzoffset int32) (uint64, error) {
```
yihuang, line 411: There seems to be no error case in this function.

x/subscription/types/cronspec.go
```diff
@@ -407,14 +408,14 @@ func roundUpItem(spec CompiledCronItem, v uint32) uint32 {
...
+	count := uint64(0)
 	for {
-		beginTime = spec.RoundUp(beginTime+1, tzoffset)
+		beginTime := spec.RoundUp(beginTime+1, tzoffset)
```
tomtau, line 414: I guess one should check for the error here from the roundup?

tomtau (changes requested):
the subscription module integration test seems to fail now: https://github.com/crypto-org-chain/chain-main/pull/449/checks?check_run_id=2276535207#step:5:889 ?

x/subscription/types/cronspec.go
```diff
@@ -384,7 +384,8 @@ func (spec CompiledCronSpec) RoundUp(ts int64, tzoffset int32) int64 {
 
 	Success:
 		// Check weekday
-		if !spec.Wday.Test(uint32(Weekday(tm.Year, tm.Month, tm.Mday))) {
+		result, error := spec.Wday.Test(uint32(Weekday(tm.Year, tm.Month, tm.Mday)))
+		if error == nil && !result {
```
leejw51crypto, line 388: Test error looks exceptional case,
how about this?
--------------------------
result, error := spec.Wday.Test(uint32(Weekday(tm.Year, tm.Month, tm.Mday)))
assert(error==nil);
 		if !result {

leejw51crypto (approved):
lgtm

x/subscription/types/cronspec.go
tomtau: Plus some "recovery": https://github.com/cosmos/cosmos-sdk/blob/master/docs/core/runtx_middleware.md ? disable the module?

PR #480, opened 2021-04-15
Problem: ibc live through export-upgrade is not tested(fix #288)

### LINKED ISSUE #288: Problem: ibc live through export-upgrade is not tested
By increasing the chain version number in chain-id, ibc state in previous chain should be able to keep working after upgrade.

linfeng-crypto:
add integration test for ibc:
update dest node's chain-id in jenesis, and send coins from src node, transfer success.


leejw51crypto (approved):
lgtm

integration_tests/test_ibc.py
```diff
@@ -210,3 +211,89 @@ def test_ibc(cluster):
...
+    )["channels"]
+    assert channels[0]["state"] == "STATE_OPEN"
+    dst_channel = channels[0]["counterparty"]["channel_id"]
+    denom_hash = (
+        hashlib.sha256(f"transfer/{dst_channel}/basecro".encode()).hexdigest().upper()
+    )
+    # update chain-id
+    cluster["ibc-1"].stop_node()
+    jenesis_data = json.loads(cluster["ibc-1"].export(0))
+    jenesis_data["chain_id"] = "ibc-2"
```
yihuang, line 240:
```suggestion
    genesis_data = json.loads(cluster["ibc-1"].export(0))
    genesis_data["chain_id"] = "ibc-2"
```

integration_tests/test_ibc.py
```diff
@@ -210,3 +211,89 @@ def test_ibc(cluster):
...
+    upgrade_time = datetime.utcnow() + timedelta(seconds=5)
+    upgrade_time.replace(tzinfo=None).isoformat("T") + "Z"
+    jenesis_data["genesis-time"] = str(upgrade_time)
+    cluster["ibc-1"].update_genesis(0, jenesis_data)
```
yihuang, line 244: - need to update genesis for both nodes?
- need to do a `unsafe-reset-all`?

integration_tests/test_ibc.py
linfeng-crypto: done

integration_tests/test_ibc.py
```diff
@@ -210,3 +211,89 @@ def test_ibc(cluster):
...
+    # update chain-id
+    cluster["ibc-1"].stop_node()
+    genesis_data = json.loads(cluster["ibc-1"].export(0))
+    genesis_data["chain_id"] = "ibc-2"
+    upgrade_time = datetime.utcnow() + timedelta(seconds=5)
+    upgrade_time.replace(tzinfo=None).isoformat("T") + "Z"
+    genesis_data["genesis-time"] = str(upgrade_time)
+    cluster["ibc-1"].update_genesis(0, genesis_data)
+    cluster["ibc-1"].start_node(0)
+    rsp = json.loads(subprocess.check_output(relayer + ["chains", "list", "--json"]))
```
yihuang, line 246: I think relayer also need to update the chain-id in config accordingly.

yihuang (commented):
I think the change on genesis file don't take effect, try to do `unsafe-reset-all` on all the nodes?

integration_tests/test_ibc.py
```diff
@@ -210,3 +211,89 @@ def test_ibc(cluster):
...
+    ]
+
+    # do a transfer from ibc-0 to ibc-2
+    balance_0 = cluster["ibc-0"].balance(cluster["ibc-0"].address("relayer"))
+    recipient = cluster["ibc-1"].address("relayer")
+    rsp = cluster["ibc-0"].ibc_transfer(
+        "relayer", recipient, "10000basecro", channels[0]["channel_id"], 1
+    )
+    assert rsp["code"] == 0, rsp["raw_log"]
+    time.sleep(10)
```
JayT106, line 276: Is it possible to detect the balance changed instead of using 10 secs sleep? We can set a 10 secs timeout to return test fail in case it really needs 10 secs to finish the transfer.

integration_tests/test_ibc.py
linfeng-crypto: in my test no need to restart relayer, it works with the old chain-id, I'll test to update the new chain-id in the relayer config and restart the relayer to see the result.

tomtau (approved):
the hardcoded timeout  can be addressed in a separate PR

PR #482, opened 2021-04-15
Problem: potential problems with the subscription module (fixes #481)

tomtau:
Solution: wrote up ADR with the proposed CosmWasm replacement approach

PR #483, opened 2021-04-20
Problem: potential problems with the subscription module

tomtau:
Solution: as per https://github.com/crypto-org-chain/chain-main/pull/482
the module is to be later replaced with an equivalent simplified
functionality in CosmWasm


leejw51crypto (approved):
lgtm

yihuang (approved):

JayT106 (approved):

PR #494, opened 2021-04-23
Problem: too long default timeout_commit value

### LINKED ISSUE #486: Problem: too long default timeout_commit value
@lezzokafka mentioned that there was some feedback that the default setting of 5s is too long that it may be good to shorten it -- but probably not too short to prevent the accidental timeouts and round increase.

devashishdxt:
Solution: changed default timeout_commit value to 3 seconds. Fixes #486.

These are the new defaults:

```
# How long we wait for a proposal block before prevoting nil
timeout_propose = "3s"
# How much timeout_propose increases with each round
timeout_propose_delta = "500ms"
# How long we wait after receiving +2/3 prevotes for “anything” (ie. not a single block or nil)
timeout_prevote = "1s"
# How much the timeout_prevote increases with each round
timeout_prevote_delta = "500ms"
# How long we wait after receiving +2/3 precommits for “anything” (ie. not a single block or nil)
timeout_precommit = "1s"
# How much the timeout_precommit increases with each round
timeout_precommit_delta = "500ms"
# How long we wait after committing a block, before starting on the new
# height (this gives us a chance to receive some more precommits, even
# though we already have +2/3).
timeout_commit = "3s"
```

### DIFF
--- a/cmd/chain-maind/app/app.go
+++ b/cmd/chain-maind/app/app.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"time"
 
 	"github.com/imdario/mergo"
 	"github.com/spf13/cast"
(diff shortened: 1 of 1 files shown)


tomtau (commented):
this appears to override it globally rather than just change the default "init" value?

devashishdxt:
> this appears to override it globally rather than just change the default "init" value?

Yeah. Ideally it should be changed here: https://github.com/cosmos/cosmos-sdk/blob/3c65c3dacded18f2d0b9f607bc7f679feff6afad/server/util.go#L202

But, that'll change the default value for all cosmos-sdk chains. I didn't find any direct way to change these defaults (other than changing the `config.toml` after `init`).

devashishdxt:
I'll try to patch the config file in `postRun` of `init` command.

devashishdxt:
> this appears to override it globally rather than just change the default "init" value?

Changed. Now we only set the default value in `PreRun` of `init` command.

tomtau (approved):

joszkomm (approved):

PR #516, opened 2021-04-29
Problem: simple NFT module not reviewed and tested

### LINKED ISSUE #507: Problem: simple NFT module not reviewed and tested
https://github.com/irisnet/irismod/tree/master/modules/nft

open a PR to vendor in the code + add basic integration tests.

Optionally, a small enhancement may be added (e.g. network params to disable/enable individual functionality -- minting, burning, transfer, edit -- or restrict to the usage to a dedicated address)

devashishdxt:
Solution: Added NFT module (code is ported from iris). Fixes #507.


tomtau (commented):
some linter issues: https://github.com/crypto-org-chain/chain-main/pull/516/checks?check_run_id=2464313189#step:4:79

tomtau:
> ❌ `/runsim` simulation test has failed 😅
> Can further check [here](https://github.com/crypto-org-chain/chain-main/actions/runs/795191737)

@allthatjazzleo

NOTICE
```diff
@@ -18,3 +18,8 @@ This project contains portions of code derived from the following libraries:
  * Copyright (c) 2014 Will Fitzgerald.
  * License: BSD 3-Clause "New" or "Revised" License
  * Repository: https://github.com/willf/bitset
+
+* irismod
+ * Copyright (c) 2016 Shanghai Bianjie AI Technology Inc.
```
tomtau, line 23:
```suggestion
 * Copyright (c) 2016-2021 Shanghai Bianjie AI Technology Inc.
```

allthatjazzleo:
> > ❌ `/runsim` simulation test has failed 😅
> > Can further check [here](https://github.com/crypto-org-chain/chain-main/actions/runs/795191737)
> 
> @allthatjazzleo
@tomtau 
`Container action is only supported on Linux`
I can update the action to JavaScript to support macOS, Windows instead
https://docs.github.com/en/actions/creating-actions/about-actions#types-of-actions

proto/nft/v1/genesis.proto
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) 2016 Shanghai Bianjie AI Technology Inc.
+// Modifications Copyright (c) 2020, Foris Limited (licensed under the Apache License, Version 2.0)
```
tomtau, line 2: plus all the other ones
```suggestion
// Copyright (c) 2016-2021 Shanghai Bianjie AI Technology Inc. (licensed under the Apache License, Version 2.0)
// Modifications Copyright (c) 2021, CRO Protocol Labs ("Crypto.org") (licensed under the Apache License, Version 2.0)
```

tomtau:
> @tomtau
> `Container action is only supported on Linux`
> I can update the action to JavaScript to support macOS, Windows instead
> https://docs.github.com/en/actions/creating-actions/about-actions#types-of-actions

@allthatjazzleo or possible to update GH action conditions here, so that the Windows build isn't trigger for the simulations?

proto/nft/v1/genesis.proto
devashishdxt: Done

allthatjazzleo:
> > @tomtau
> > `Container action is only supported on Linux`
> > I can update the action to JavaScript to support macOS, Windows instead
> > https://docs.github.com/en/actions/creating-actions/about-actions#types-of-actions
> 
> @allthatjazzleo or possible to update GH action conditions here, so that the Windows build isn't trigger for the simulations?

yes. I think it is better solution.
in cosmos sdk, it also run [simulations](https://github.com/cosmos/cosmos-sdk/blob/master/.github/workflows/sims.yml) on linux only.

x/nft/exported/nft.go
```diff
@@ -10,7 +10,7 @@ import (
 type NFT interface {
 	GetID() string
 	GetName() string
-	GetOwner() sdk.AccAddress
+	GetOwner() (sdk.AccAddress, error)
```
tomtau, line 13: when would `GetOwner` return error?

x/nft/keeper/collection.go
```diff
@@ -18,14 +18,20 @@ import (
 // SetCollection saves all NFTs and returns an error if there already exists
```
tomtau, line 18: not sure if the comment should be updated -- before, it returned error if any of the NFTs existed (i.e. `k.MintNFT` returned an err)... now, also if `nft.GetOwner` returns an error

x/nft/exported/nft.go
devashishdxt: https://github.com/crypto-org-chain/chain-main/pull/516/files#diff-56c6844365f0697399efd4332313e84170925e99bd1479be9a9403dbd3e7d564R37

x/nft/keeper/msg_server.go
```diff
@@ -51,6 +51,7 @@ func (m msgServer) IssueDenom(goCtx context.Context, msg *types.MsgIssueDenom) (
 	return &types.MsgIssueDenomResponse{}, nil
 }
 
+// nolint: dupl
```
tomtau, line 54: perhaps a TODO to refactor this if it'll make it better -- I guess the linter error is about the nearly duplicate "EmitEvents" bit which could perhaps be abstracted out?

x/nft/exported/nft.go
devashishdxt: Current code assumes that the bech32 owner address is valid and discards the error when converting bech32 address into `sdk.AccAddress`.

x/nft/keeper/collection.go
devashishdxt: Done

yihuang (approved):

x/nft/keeper/msg_server.go
devashishdxt: Yeah. Because the `msg` is of different type with almost similar fields (but they do not implement same interface). I'm not sure how we can abstract this out. Is there something similar to Rust macros in Go?

I'll add a `TODO` for now.

x/nft/exported/nft.go
tomtau: I looked down and it seems the error will only be if `SplitKeyOwner` returns an error... which would only be in the case of a corrupted storage? if so, perhaps just panic there instead of bubbling this up (panics should be caught by the Cosmos SDK middleware afaik)?

x/nft/keeper/msg_server.go
devashishdxt: Done

x/nft/keeper/querier.go
```diff
@@ -101,7 +106,11 @@ func queryDenom(ctx sdk.Context, req abci.RequestQuery, k Keeper, legacyQuerierC
 		return nil, sdkerrors.Wrap(sdkerrors.ErrJSONUnmarshal, err.Error())
 	}
 
-	denom, _ := k.GetDenom(ctx, params.ID)
+	denom, err := k.GetDenom(ctx, params.ID)
+
+	if err != nil {
+		return nil, err
```
tomtau, line 112: should this do some `sdkerrors.Wrap(sdkerrors.<some error, e.g. invalid request>, err.Error())`?

devashishdxt: `GetDenom` already returns wrapped error

x/nft/exported/nft.go
devashishdxt: Ok

x/nft/types/denom.go
```diff
@@ -7,6 +7,7 @@ import (
 )
 
 // NewDenom return a new denom
+// nolint: interfacer
```
tomtau, line 10: false positive?

x/nft/types/genesis.go
```diff
@@ -22,7 +22,13 @@ func ValidateGenesis(data GenesisState) error {
 		}
 
 		for _, nft := range c.NFTs {
-			if nft.GetOwner().Empty() {
+			owner, err := nft.GetOwner()
+
+			if err != nil {
+				return err
```
tomtau, line 28: if the error is kept, `sdkerrors.Wrap`

x/nft/types/nft.go
```diff
@@ -11,6 +11,7 @@ import (
 var _ exported.NFT = BaseNFT{}
 
 // NewBaseNFT creates a new NFT instance
+// nolint: interfacer
```
tomtau, line 14: false positive?

x/nft/types/nft.go
```diff
@@ -32,9 +33,8 @@ func (bnft BaseNFT) GetName() string {
 }
 
 // GetOwner return the owner of BaseNFT
-func (bnft BaseNFT) GetOwner() sdk.AccAddress {
-	owner, _ := sdk.AccAddressFromBech32(bnft.Owner)
-	return owner
+func (bnft BaseNFT) GetOwner() (sdk.AccAddress, error) {
```
tomtau, line 36: does the BaseNFT.owner need to be "string" -- could it be `sdk.AccAddress` instead?

x/nft/exported/nft.go
tomtau: or is this interface ever called on non-validated user input?

x/nft/types/nft.go
tomtau: I guess it's from protobuf, so needs string?

x/nft/exported/nft.go
tomtau: @yihuang @leejw51crypto any thoughts on this? This would be good to have a second pair of eyes

x/nft/types/nft.go
devashishdxt: Yes. This is from protobuf so it needs to be a string.

proto/nft/v1/tx.proto
```diff
@@ -0,0 +1,99 @@
...
+// Modifications Copyright (c) 2021, CRO Protocol Labs ("Crypto.org") (licensed under the Apache License, Version 2.0)
+syntax = "proto3";
+package chainmain.nft.v1;
+
+import "gogoproto/gogo.proto";
+
+option go_package = "github.com/crypto-org-chain/chain-main/x/nft/types";
+option (gogoproto.goproto_getters_all) = false;
+
+// Msg defines the htlc Msg service.
```
tomtau, line 11:
```suggestion
// Msg defines the NFT Msg service.
```

proto/nft/v1/tx.proto
```diff
@@ -0,0 +1,99 @@
...
+service Msg {
+    // IssueDenom defines a method for issue a denom.
+    rpc IssueDenom(MsgIssueDenom) returns (MsgIssueDenomResponse);
+
+    // MintNFT defines a method for mint a new nft
+    rpc MintNFT(MsgMintNFT) returns (MsgMintNFTResponse);
+
+    // RefundHTLC defines a method for editing a nft.
```
tomtau, line 19:
```suggestion
    // EditNFT defines a method for editing a nft.
```

proto/nft/v1/tx.proto
```diff
@@ -0,0 +1,99 @@
...
+import "gogoproto/gogo.proto";
+
+option go_package = "github.com/crypto-org-chain/chain-main/x/nft/types";
+option (gogoproto.goproto_getters_all) = false;
+
+// Msg defines the htlc Msg service.
+service Msg {
+    // IssueDenom defines a method for issue a denom.
+    rpc IssueDenom(MsgIssueDenom) returns (MsgIssueDenomResponse);
+
```
tomtau, line 15: we can probably keep the names, but just may need some comments to note that "Denom == Collection" in here?

x/nft/spec/01_state.md
```diff
@@ -0,0 +1,54 @@
...
+Nft defines the tokenData of non-fungible tokens, mainly including ID, owner, and tokenURI.Nft can be transferred through `MsgTransferNFT`, or you can edit `tokenURI` information through `MsgEditNFT` transaction. The name of the collection and the id of nft identify the unique assets in the system. The `NFT` Interface inherits the BaseNFT struct and includes getter functions for the asset data. It also includes a Stringer function in order to print the struct. The interface may change if tokenData is moved to it’s own module as it might no longer be necessary for the flexibility of an interface.
+
+```go
+// NFT non fungible token interface
+type NFT interface {
+    GetID() string              // unique identifier of the NFT
+    GetName() string            // return the name of BaseNFT
+    GetOwner() sdk.AccAddress   // gets owner account of the NFT
+    GetURI() string             // tokenData field: URI to retrieve the of chain tokenData of the NFT
+    GetData() string            // return the Data of BaseNFT
```
tomtau, line 17: No size limits on Name, ID, Data...?

x/nft/exported/nft.go
devashishdxt: There are two functions named `GetOwner` in the module. This `GetOwner` does not call `SplitKeyOwner`. It only parses bech32 address. Even if we find that this function is never called on non-validated user input, this'll not guarantee future changes to the code will not do that. We can decide between two options:

- Returning `error`
- panic

@yihuang @leejw51crypto @linfeng-crypto @JayT106 any comments?

x/nft/spec/02_messages.md
```diff
@@ -0,0 +1,122 @@
...
+    Id     string
+    Name   string
+    Schema string
+    Sender string
+}
+```
+
+## MsgTransferNFT
+
+This is the most commonly expected MsgType to be supported across chains. While each application specific blockchain will have very different adoption of the `MsgMintNFT`, `MsgBurnNFT` and `MsgEditNFT` it should be expected that most chains support the ability to transfer ownership of the NFT asset. The exception to this would be non-transferable NFTs that might be attached to reputation or some asset which should not be transferable. It still makes sense for this to be represented as an NFT because there are common queriers which will remain relevant to the NFT type even if non-transferable. This Message will fail if the NFT does not exist. By default it will not fail if the transfer is executed by someone beside the owner. **It is highly recommended that a custom handler is made to restrict use of this Message type to prevent unintended use.**
```
tomtau, line 28: > By default it will not fail if the transfer is executed by someone beside the owner. **It is highly recommended that a custom handler is made to restrict use of this Message type to prevent unintended use.**

@devashishdxt @yihuang @leejw51crypto

x/nft/keeper/keeper.go
```diff
@@ -0,0 +1,152 @@
...
+		nft.Name = tokenNm
+	}
+
+	if types.Modified(tokenURI) {
+		nft.URI = tokenURI
+	}
+
+	if types.Modified(tokenData) {
+		nft.Data = tokenData
+	}
```
tomtau, line 129: ummm... so each transfer can potentially change the NFT content?

tomtau (commented):
Unless I missed something, it seems the default keeper handlers have very relaxed handling of message processing -- e.g. one can modify the NFT content any time... which probably shouldn't be allowed?

x/nft/spec/02_messages.md
devashishdxt: @tomtau Is it possible to create issues for these things and do them in separate PRs. If we include all the changes in one PR, it'll become too big (which it already is) and it'll be difficult to review too.

We can keep the main issue open and add sub-issues to it.

devashishdxt:
@tomtau Is it possible to create issues for extra decisions (other than porting the iris code to chain-main) and do them in separate PRs. If we include all the changes in one PR, it'll become too big (which it already is) and it'll be difficult to review too.

We can keep the main issue open and add sub-issues to it.

x/nft/spec/02_messages.md
yihuang: Somehow, in the code, it do check the sender is owner though.

x/nft/exported/nft.go
yihuang: Currently `BaseNFT` is used in the `Collection` type which is part of genesis state. Genesis state I think should be considered non-validated user input? it's supposed to be validated in `ValidateGenesis`.

tomtau: I'd go with how this is done in Cosmos SDK or other modules for consistency

tomtau:
> @tomtau Is it possible to create issues for extra decisions (other than porting the iris code to chain-main) and do them in separate PRs. If we include all the changes in one PR, it'll become too big (which it already is) and it'll be difficult to review too.
> 
> We can keep the main issue open and add sub-issues to it.

@devashishdxt yeah, I'll update the ADR and create corresponding issues

x/nft/exported/nft.go
leejw51crypto: agree, for corrupted storage, panic make sense

leejw51crypto (approved):
lgtm

x/nft/types/expected_keepers.go
```diff
@@ -0,0 +1,22 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+)
+
+// AccountKeeper defines the expected account keeper for query account
+type AccountKeeper interface {
+	GetAccount(ctx sdk.Context, addr sdk.AccAddress) authtypes.AccountI
+}
+
+// BankKeeper defines the expected interface needed to retrieve account balances.
```
tomtau, line 15: I guess this only needed for simulations?

tomtau (commented):
should sim_test.go be also updated (given the NFT module contains its simulation operations)?

x/nft/client/testutil/test_helpers.go
```diff
@@ -0,0 +1,132 @@
...
+
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/cosmos/cosmos-sdk/testutil"
+	clitestutil "github.com/cosmos/cosmos-sdk/testutil/cli"
+
+	nftcli "github.com/crypto-org-chain/chain-main/v2/x/nft/client/cli"
+)
+
+// MsgRedelegateExec creates a redelegate message.
```
tomtau, line 18: `IssueDenomExec`

x/nft/exported/nft.go
JayT106: agree to use panic, there is an example for handling the error after calling `AccAddressFromBech32`
https://github.com/cosmos/cosmos-sdk/blob/be4a965599dfae6992451e4b5135c487ed45053b/x/bank/types/balance.go#L17

JayT106 (approved):

x/nft/exported/nft.go
devashishdxt: Done

x/nft/types/genesis.go
devashishdxt: Error was removed

proto/nft/v1/tx.proto
devashishdxt: Done

proto/nft/v1/tx.proto
devashishdxt: Done

x/nft/types/expected_keepers.go
devashishdxt: This is also used in `NewAppModule()`: https://github.com/crypto-org-chain/chain-main/pull/516/files/f828f433f2c80213a04436c6dc999e6e3809e5fd#diff-0fd2887ee17fe532655e8bd2fe7bc4ba896849886b7492dcb0b76824bbb5e95bR106

x/nft/client/testutil/test_helpers.go
devashishdxt: Done

x/nft/types/denom.go
devashishdxt: I get this lint error when I run `golangci-lint` locally. Anyway, `interfacer` is now deprecated (https://github.com/mvdan/interfacer). Should we remove this from github ci config?

x/nft/types/nft.go
devashishdxt: Same as above

devashishdxt:
> should sim_test.go be also updated (given the NFT module contains its simulation operations)?

Yes. I'll change.

devashishdxt:
> should sim_test.go be also updated (given the NFT module contains its simulation operations)?

Done

x/nft/keeper/keeper.go
devashishdxt: Yeah. This is weird. Ideally, `TransferNFT` should only take `nftId` and `toAddress`? And it should only make the transfer if the message is signed by the owner of the NFT?

PR #529, opened 2021-05-05
Problem: Vendored NFT module's permissions are too lax

### LINKED ISSUE #520: Problem: vendored NFT module's permissions are too lax
> By default it will not fail if the transfer is executed by someone beside the owner. It is highly recommended that a custom handler is made to restrict use of this Message type to prevent unintended use.

The code seems to check for the owner==sender, but it allows the owner to modify the token content:

this should be modified to match similar business logic to CW721 (only the token issuer can edit and burn its tokens; the token owner can only transfer): spec + implementation need to be updated + integration tests for this expected logic

devashishdxt:
Solution: Changed the implementation as per new specs. Fixes #520.


proto/nft/v1/tx.proto
```diff
@@ -45,11 +45,8 @@ message MsgTransferNFT {
 
     string id = 1;
     string denom_id = 2 [ (gogoproto.moretags) = "yaml:\"denom_id\"" ];
-    string name = 3;
-    string uri = 4 [ (gogoproto.customname) = "URI" ];
-    string data = 5;
-    string sender = 6;
-    string recipient = 7;
+    string sender = 3;
+    string recipient = 4;
```
tomtau, line 49: not a protobuf expert -- if this was meant to be still somewhat compatible with the original IRISnet module... what's the best way when the fields are removed?

tomtau: (if possible, not a major blocker)

x/nft/keeper/denom.go
```diff
@@ -64,3 +64,21 @@ func (k Keeper) GetDenoms(ctx sdk.Context) (denoms []types.Denom) {
...
+// nolint: interfacer
+func (k Keeper) IsDenomCreator(ctx sdk.Context, denomID string, address sdk.AccAddress) (types.Denom, error) {
+	denom, err := k.GetDenom(ctx, denomID)
+	if err != nil {
+		return types.Denom{}, err
+	}
+
+	addressStr := address.String()
+
+	if addressStr != denom.Creator {
```
tomtau, line 79: I'd probably add a comment here why the string comparison is ok -- is the denom.creator guaranteed/checked to be a valid address or is it possible one can set it to invalid string (empty, some random...)?

brianatcrypto (approved):
changes look sensible.

tomtau (approved):

proto/nft/v1/tx.proto
devashishdxt: not changing the index?

x/nft/keeper/denom.go
devashishdxt: Changed to comparing concrete address types

proto/nft/v1/tx.proto
yihuang: with protobuf rules, if you want to keep compatibility, never remove fields

tomtau: maybe no need -- the end goal will be to migrate to standard Cosmos SDK module... for compat with IRISnet -- the response data didn't change, so browsers/wallets should be able to query / view the same things?

yihuang (approved):

tomtau (approved):

matias904 (approved):

PR #542, opened 2021-05-10
Problem: DenomID in NFT module does not accept capital letters

devashishdxt:
Solution: Modified error message clarifying that DenomID only accepts lowecase letters. Fixes #542.

### DIFF
--- a/x/nft/types/validation.go
+++ b/x/nft/types/validation.go
@@ -50,7 +50,7 @@ func ValidateTokenID(tokenID string) error {
 		return sdkerrors.Wrapf(ErrInvalidTokenID, "the length of nft id(%s) only accepts value [%d, %d]", tokenID, MinDenomLen, MaxDenomLen)
 	}
 	if !IsBeginWithAlpha(tokenID) || !IsAlphaNumeric(tokenID) {
-		return sdkerrors.Wrapf(ErrInvalidTokenID, "nft id(%s) only accepts lowecase alphanumeric characters, and begin with an english letter", tokenID)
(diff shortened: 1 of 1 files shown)


tomtau (commented):
is that a problem -- or is it better for denoms to be lowercase only (to prevent confusion or similarly named denoms)?
and if so, whether the testnet should go with the current 2.0.0 pre-release or create a new release with this patch?

leejw51crypto:
small123, SMALL123,
Small123 all will be different denomID, that can cause some confusion

devashishdxt:
> is that a problem -- or is it better for denoms to be lowercase only (to prevent confusion or similarly named denoms)?
> and if so, whether the testnet should go with the current 2.0.0 pre-release or create a new release with this patch?

We can use small letters only. In that case we can modify the error message clarifying this?

devashishdxt:
> yes, modifying error message seems better solution

Changed

leejw51crypto (approved):
lgtm,

small spelling issue in nft-id

x/nft/types/validation.go
```diff
@@ -50,7 +50,7 @@ func ValidateTokenID(tokenID string) error {
 		return sdkerrors.Wrapf(ErrInvalidTokenID, "the length of nft id(%s) only accepts value [%d, %d]", tokenID, MinDenomLen, MaxDenomLen)
 	}
 	if !IsBeginWithAlpha(tokenID) || !IsAlphaNumeric(tokenID) {
-		return sdkerrors.Wrapf(ErrInvalidTokenID, "nft id(%s) only accepts alphanumeric characters, and begin with an english letter", tokenID)
+		return sdkerrors.Wrapf(ErrInvalidTokenID, "nft id(%s) only accepts lowecase alphanumeric characters, and begin with an english letter", tokenID)
```
leejw51crypto, line 53: lowercase?

PR #548, opened 2021-05-14
Problem: goreleaser config is not correct

tomtau:
Solution:
- update the config based on 2b42e746327be0817a79e7f04894e2bde888ddfb
without the testnet


yihuang (commented):
Need to add `archives.builds` too?

tomtau:
> Need to add `archives.builds` too?

@yihuang a good catch!

yihuang (approved):

PR #552, opened 2021-05-17
Problem: nft module store is not initialized after upgrade

yihuang:
Solution:
- call SetStoreLoader to initialise nft module store

👮🏻👮🏻👮🏻 !!!! REFERENCE THE PROBLEM YOUR ARE SOLVING IN THE PR TITLE AND DESCRIBE YOUR SOLUTION HERE !!!! DO NOT FORGET !!!! 👮🏻👮🏻👮🏻

# PR Checklist:

- [ ] Have you read the [CONTRIBUTING.md](https://github.com/crypto-org-chain/chain-main/blob/master/CONTRIBUTING.md)?
- [ ] Have you rebased your work on top of the latest master? 
- [ ] Have you checked your code compiles? (`make`)
- [ ] Have you included tests for any non-trivial functionality?
- [ ] If you are contributing for the first time, please read the agreement in [CONTRIBUTING.md](https://github.com/crypto-org-chain/chain-main/blob/master/CONTRIBUTING.md) now and add a comment to this pull request stating that your PR is in accordance with the [Developer's Certificate of Origin](https://github.com/crypto-org-chain/chain-main/blob/master/CONTRIBUTING.md#developer-certificate-of-originn).

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -26,6 +26,7 @@ import (
 	"github.com/cosmos/cosmos-sdk/server/api"
 	"github.com/cosmos/cosmos-sdk/server/config"
 	servertypes "github.com/cosmos/cosmos-sdk/server/types"
+	storetypes "github.com/cosmos/cosmos-sdk/store/types"
 	"github.com/cosmos/cosmos-sdk/testutil/testdata"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/types/module"
@@ -414,6 +415,27 @@ func New(
 	)
 	app.SetEndBlocker(app.EndBlocker)
 
+	planName := "v2.0.0"
+	app.UpgradeKeeper.SetUpgradeHandler(planName, func(ctx sdk.Context, _ upgradetypes.Plan) {
+		// https://github.com/crypto-org-chain/chain-main/blob/master/doc/architecture/adr-003.md
+	})
+
+	upgradeInfo, err := app.UpgradeKeeper.ReadUpgradeInfoFromDisk()
+	if err != nil {
+		panic(err)
+	}
+
+	if upgradeInfo.Name == planName && !app.UpgradeKeeper.IsSkipHeight(upgradeInfo.Height) {
+		storeUpgrades := storetypes.StoreUpgrades{
+			Added:   []string{nfttypes.StoreKey},
+			Renamed: []storetypes.StoreRename{},
+			Deleted: []string{},
+		}
+
+		// configure store loader that checks if version == upgradeHeight and applies store upgrades
+		app.SetStoreLoader(upgradetypes.UpgradeStoreLoader(upgradeInfo.Height, &storeUpgrades))
+	}
+
 	if loadLatest {
 		if err := app.LoadLatestVersion(); err != nil {
 			tmos.Exit(err.Error())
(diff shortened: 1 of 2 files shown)


tomtau (approved):

app/app.go
```diff
@@ -414,6 +415,27 @@ func New(
...
+
+	if upgradeInfo.Name == planName && !app.UpgradeKeeper.IsSkipHeight(upgradeInfo.Height) {
+		storeUpgrades := storetypes.StoreUpgrades{
+			Added:   []string{nfttypes.StoreKey},
+			Renamed: []storetypes.StoreRename{},
+			Deleted: []string{},
+		}
+
+		// configure store loader that checks if version == upgradeHeight and applies store upgrades
+		app.SetStoreLoader(upgradetypes.UpgradeStoreLoader(upgradeInfo.Height, &storeUpgrades))
```
tomtau, line 436: should this just be called once / inside `SetUpgradeHandler` ?

devashishdxt (approved):

app/app.go
```diff
@@ -414,6 +415,27 @@ func New(
 	)
 	app.SetEndBlocker(app.EndBlocker)
 
+	planName := "v2.0.0"
```
leejw51crypto, line 418: can this fetched from other place?
because two places hard-coded

app/app.go
```diff
@@ -414,6 +415,27 @@ func New(
...
+	app.UpgradeKeeper.SetUpgradeHandler(planName, func(ctx sdk.Context, _ upgradetypes.Plan) {
+		// https://github.com/crypto-org-chain/chain-main/blob/master/doc/architecture/adr-003.md
+	})
+
+	upgradeInfo, err := app.UpgradeKeeper.ReadUpgradeInfoFromDisk()
+	if err != nil {
+		panic(err)
+	}
+
+	if upgradeInfo.Name == planName && !app.UpgradeKeeper.IsSkipHeight(upgradeInfo.Height) {
```
leejw51crypto, line 428: could add some explain for IsSkipHeight,
such as height > upgradeInfo.Height?

leejw51crypto: in latest planName is different from "2.0.0"? 
and not fall into this step?

app/app.go
yihuang: I think the override only works before storage constructed, so it need to happen pretty early. while upgrade handler is called in abci event.

yihuang: I get the example code from the godoc of upgrade module: https://pkg.go.dev/github.com/cosmos/cosmos-sdk@v0.42.4/x/upgrade

app/app.go
yihuang: I think the plan name only specified once here.

app/app.go
yihuang: https://github.com/cosmos/cosmos-sdk/blob/master/x/upgrade/types/storeloader.go#L13
And before apply the upgrade, it'll check the height again, so it should be safe here.

app/app.go
yihuang: cosmos-sdk has a feature that you can skip upgrades in command line:
```
$ chain-maind start --help
...
      --unsafe-skip-upgrades ints                       Skip a set of upgrade heights to continue the old binary
```

leejw51crypto (approved):
lgtm

Issue #278, opened 2020-11-18
Problem: incorrect codecov reporting?


tomtau:
In https://codecov.io/gh/crypto-com/chain-main/tree/master/x/supply/keeper
it shows that only 13 lines are hit in querier.go
but with this test: https://github.com/crypto-com/chain-main/blob/master/integration_tests/test_basic.py#L64 -- it should be more (imports + `NewQuerier` path to types.QueryLiquidSupply + `queryLiquidSupply`) or does that test cover something else?

yihuang:
```
$ go test -v -coverprofile=size_coverage.out ./... && go tool cover -html=size_coverage.out
```
don't cover the `New` function in `app/app.go`. It's called in some unit tests like `TestSubscriptionCollect`.

tomtau:
@calvinaco 's codecov setup suggestion:
```
go test -race -coverprofile=coverage.txt -covermode=atomic
```

yihuang:
> @calvinaco 's codecov setup suggestion:
> 
> ```
> go test -race -coverprofile=coverage.txt -covermode=atomic
> ```

I just tried this, didn't help.

CeruleanAtMonaco:
upvote to fix this issue, accurate test coverage would help us to spot where the real weak points are

------details-----
as of commit `bf3ce0d52c169d0780cf050e3bbb86dab07b30a1` when running `make test`

github.com/crypto-org-chain/chain-main/v1/app	0.930s	coverage: 1.2% of statements
github.com/crypto-org-chain/chain-main/v1/config	0.437s	coverage: 83.3% of statements
github.com/crypto-org-chain/chain-main/v1/x/subscription/keeper	0.913s	coverage: 53.5% of statements
github.com/crypto-org-chain/chain-main/v1/x/subscription/types	0.469s	coverage: 6.1% of statements

PR #578, opened 2021-06-09
Problem: integration test takes too long time to complete (fix #489)

### LINKED ISSUE #489: Problem: integration test takes too long time to complete
how about split into 2 or 3 yml
for example,
build1.yml, build2.yml, build3.yml
or some large category

and run simultaneously

leejw51crypto:
Solution: use concurrent jobs using pytest.marks

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -6,58 +6,149 @@ on:
       - master
       - release/**
     tags:
-      - '*'
+      - "*"
 
 jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
-      with:
+      - uses: actions/checkout@v2
+        with:
           submodules: true
-    - uses: cachix/install-nix-action@v12
-    - uses: cachix/cachix-action@v8
-      with:
-        name: crypto-com
-        skipPush: true
-    - name: lint
-      run: make lint-ci
+      - uses: cachix/install-nix-action@v12
+      - uses: cachix/cachix-action@v8
+        with:
+          name: crypto-com
+          skipPush: true
+      - name: lint
+        run: make lint-ci
 
   test:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
-      with:
+      - uses: actions/checkout@v2
+        with:
           submodules: true
-    - uses: cachix/install-nix-action@v12
-    - uses: cachix/cachix-action@v8
-      with:
-        name: crypto-com
-        signingKey: '${{ secrets.CACHIX_SIGNING_KEY }}'
-        skipPush: ${{github.event_name == 'pull_request'}}
-    - name: Run integration tests
-      run: make nix-integration-test
-    - name: Build chain utils
-      run: nix-build -A chain-utils-testnet
-    - name: Upload coverage report
-      uses: codecov/codecov-action@v1
-      with:
-        token: ${{ secrets.CODECOV_TOKEN }}
-        file: ./coverage.*.txt
-        flags: integration_tests
-    - name: Publish docker image
-      if: github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/')
-      run: |
-        nix-build -A pystarportImage docker.nix
-        IMAGE=$(docker load < result)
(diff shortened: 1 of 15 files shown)


leejw51crypto:
031FE4AC-EB7D-42DD-BF90-F65C0DF7C39E

default.nix
```diff
@@ -133,9 +133,7 @@ rec {
     set -e
     export PATH=${ci-env}/bin:$PATH
     export TESTS=${tests_src}/integration_tests
-    pytest -v -n 3 -m 'not ledger and not upgrade' --dist loadscope $TESTS
-    pytest -v -m upgrade $TESTS
-    pytest -v -m ledger $TESTS
+    $1 $TESTS
```
yihuang, line 136:
```suggestion
    CMD=${1:-"pytest -v"}
    $1 $TESTS
```

in case someone `run-integration-tests` directly.

tomtau (commented):
seems ok, but noticed that the codecov decreased -- meaning that the instrumentation or collecting the coverage results changed...
the coverage from the integration had issues anyway though: https://github.com/crypto-org-chain/chain-main/issues/278

yihuang (commented):
are grpc tests missed?

.github/workflows/nix.yml
```diff
@@ -6,58 +6,169 @@ on:
...
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: true
+      - uses: cachix/install-nix-action@v12
+      - uses: cachix/cachix-action@v8
+        with:
+          name: crypto-com
+          signingKey: "${{ secrets.CACHIX_SIGNING_KEY }}"
+          skipPush: ${{github.event_name == 'pull_request'}}
```
yihuang, line 76:
```suggestion
```

These two lines can be removed for the other `tests-*` tasks, since we only need to push once in the default test job.

.github/workflows/nix.yml
```diff
@@ -6,58 +6,169 @@ on:
...
+      - name: Run integration tests
+        run: make nix-integration-test-gov
+
+  test-coverage:
```
yihuang, line 155: What do this job do? I suppose there's no coverage files in this job to collect?

yihuang (commented):
`test_staking_unbond` is run twice, since it have both `normal` and `slow` mark, maybe remove the `slow` mark.

yihuang (changes requested):
- some test cases (`test_staking_unbond`) are run multiple times since they have multiple marks, better remove one of the mark.
- grpc test cases are not included?
- test-coverage can be removed?
- `skipPush` better be `false` in `tests-*` jobs, since the `tests` job will do the push, so other jobs don't have to. (when `signingKey` not set, it's automatically `false`)

.github/workflows/nix.yml
leejw51crypto: ok

leejw51crypto:
ok, i'll modify for above

yihuang:
> ok, i'll modify for above

to fix the lint issue, you might need to escape the dollar sign by using `$$`.

Makefile
```diff
@@ -153,7 +153,29 @@ make-proto:
 ###############################################################################
 # nix installation: https://nixos.org/download.html
 nix-integration-test: check-network make-proto
-	nix run -f ./. run-integration-tests -c run-integration-tests
+	nix run -f ./default.nix run-integration-tests -c run-integration-tests "pytest -v -m normal"
+
+nix-integration-test-upgrade: check-network make-proto
```
yihuang, line 158:
```suggestion
nix-integration-test-upgrade: check-network
```

Now only the `nix-integration-test-grpc` need the `make-proto` dependency, right?

yihuang (approved):

.github/workflows/nix.yml
leejw51crypto: removed

yihuang:
Seems good now, can you squash the commits?

devashishdxt (approved):

devashishdxt:
> Seems good now, can you squash the commits?

Can't we squash when merging?

leejw51crypto:
sure, i'm also adding all tests entry point

PR #585, opened 2021-06-23
Problem: nix function called with an unexpected argument (fixes #584)

### LINKED ISSUE #584: Problem: nix function called with an unexpected argument
https://github.com/crypto-org-chain/chain-main/runs/2881839518#step:8:1
```
error: anonymous function at /home/runner/work/chain-main/chain-main/pystarport/default.nix:1:1 called with unexpected argument 'chaind', at /home/runner/work/chain-main/chain-main/default.nix:72:16
19
(use '--show-trace' to show detailed location information)
```

tomtau:
Solution: removed the argument from the call
(argument was removed from the function in ea92405d18902e0e325f7f623eb4458b88d643f9 )

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -69,14 +69,12 @@ rec {
 
   chain-maind = build-chain-maind { };
 
-  pystarport = import ./pystarport { inherit pkgs; chaind = "${chain-maind}/bin/chain-maind"; };
+  pystarport = import ./pystarport { inherit pkgs; };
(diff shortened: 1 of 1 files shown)


Crytochain (commented):
can you also remove the `pystarport-unbind`, it's also irrelevant now.

yihuang (approved):

PR #598, opened 2021-07-13
Problem: no test for a different fee payer (Fix #591)

### LINKED ISSUE #591: Problem: no test for a different fee payer
Right now, all the integration tests have the default setup where the transaction fee is paid by the transaction sender.
The setup could be more flexible: https://github.com/cosmos/cosmos-sdk/blob/v0.42.6/x/auth/tx/builder.go#L132 where one can set another third party in `AuthInfo.Fee.Payer` of the transaction -- it'll be good to have a test that does it and verifies that the fee was deducted from the third party instead of the sender.

MaCong-crypto:
Fix #591 

Solution: add fee paid by third party in a transaction


tomtau (changes requested):
congrats on the first PR!🎆
there are some small lint issues: https://github.com/crypto-org-chain/chain-main/pull/598/checks?check_run_id=3055963168#step:5:16 but I think it should be fixable by just running "black" on those files?

integration_tests/test_fee_payer.py
```diff
@@ -0,0 +1,48 @@
...
+
+    unsigned_tx_msg = cluster.transfer(
+        sender_addr,
+        receiver_addr,
+        "%sbasecro" % transaction_coins,
+        generate_only=True,
+        fees="%sbasecro" % fee_coins,
+    )
+
+    json.dump(unsigned_tx_msg, unsigned_tx_txt.open("w"))
```
yihuang, line 33: Could have buffered file writing here, since the file object might not be closed timely, depend on gc.

integration_tests/logger.py
```diff
@@ -0,0 +1,19 @@
+import logging
```
yihuang, line 1: This module seems not used in other places?
Currently the other test cases just output stuff to stdout directly through `print`.

yihuang (changes requested):
- several small suggestions
- lint issues can be fixed by run `isort` and `black` on the source file.

integration_tests/logger.py
MaCong-crypto: Yes, this module is not used anywhere, but it is quite helpful in debugging process, as it output logs to a file by using `logger.info()`.
Problem I found with stdout is no output is printed/logged if test passed; and the output is too long if test failed currently.
I can add a small documentation inside the file to explain how to use it.

yihuang: there are several pytest options to control what's output:
- `-s`, don't capture stdout, so all printed stuff is shown
- `--supervisord-quiet`, by default the devnet logs are also redirected to stdout, with this option, they don't, you can still check the log files though, you can find them in `/tm/pytest-xxx`

MaCong-crypto: Thank you very much for the tips! 😆  I will remove this file.

tomtau (approved):

yihuang (approved):

PR #617, opened 2021-08-11
Problem: final release of SDK 0.43.0 is not used

tomtau:
Solution: upgraded the cosmos sdk and ibc-go dependencies


yihuang (approved):

tomtau:
> I'm not sure if we can do anything about the vulnerabilities in this PR.

there may be false positives -- I opened: https://github.com/crypto-org-chain/chain-main/issues/618

PR #625, opened 2021-08-16
test fee grant and revoke grant (Fix #614)

### LINKED ISSUE #614: Problem: feegrant is not tested
@MaCong-crypto https://docs.cosmos.network/master/modules/feegrant/

MaCong-crypto:
Solution:
Add following tests to test `tx feegrant grant` and `tx feegrant revoke`
- check whether granter pays transaction fee with no fee limit or grant expiry
- check transaction should fail when exceeding grant fee limit
- check transaction should fail after grant expiry date
- check whether granter pays transaction fee under period limit during each period
- check exceeding period limit should not affect the next period
- check transaction should fail if fee grant is revoked

Fix #614

### DIFF
--- a/integration_tests/test_feegrant.py
+++ b/integration_tests/test_feegrant.py
@@ -1,0 +1,270 @@
+import datetime
+from time import sleep
+
+import pytest
+
+from .utils import (
+    BASECRO_DENOM,
+    SUCCESS_CODE,
+    grant_fee_allowance,
+    revoke_fee_grant,
+    transfer,
+)
+
+pytestmark = pytest.mark.normal
+
+
+def test_basic_fee_allowance(cluster):
+    """
+    check basic fee allowance with no limit or grant expiry
+    """
+    transaction_coins = 100
+    fee_coins = 10
+
+    fee_granter_address = cluster.address("community")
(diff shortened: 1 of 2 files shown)


integration_tests/test_feegrant.py
```diff
@@ -0,0 +1,272 @@
...
+    fee_coins = 10
+
+    fee_granter_address = cluster.address("community")
+    fee_grantee_address = cluster.address("ecosystem")
+    receiver_address = cluster.address("reserve")
+
+    revoke_fee_grant(cluster, fee_granter_address, fee_grantee_address)
+    grant_fee_allowance(cluster, fee_granter_address, fee_grantee_address)
+
+    revoke_fee_grant(cluster, fee_granter_address, fee_grantee_address)
```
tomtau, line 261: why is there this double revoke?

tomtau (approved):
looks good -- I think it tests all feegrant options, well done!

integration_tests/test_feegrant.py
MaCong-crypto: All test cases execute sequentially but with different feegrant options, the first revoke before the grant removes the previous grant in previous test cases.

Issue #626, opened 2021-08-18
Problem: test_periodic_fee_allowance occasionally fails on GH Actions


tomtau:
https://github.com/crypto-org-chain/chain-main/runs/3347062763#step:5:850

MaCong-crypto:
Also need to change `test_exceed_period_limit_should_not_affect_the_next_period`

PR #630, opened 2021-08-23
test authz module (Fix #615)

### LINKED ISSUE #615: Problem: authz is not tested
@MaCong-crypto https://docs.cosmos.network/master/modules/authz/

MaCong-crypto:
Solution:
Add following tests to test `tx authz grant`, `tx authz exec` and `tx authz revoke`

- check send execution should succeed when amount is within authorization spend limit
- check send execution should fail as send amount is beyond send authorization spend limit
- check send execution should fail as grant is revoked
- check send execution should succeed when grant a generic authorization with send msg type
- check delegation execution to allowed validators in granted authorization should succeed
- check delegation execution to deny validators in granted authorization should fail
- check delegation, unbond and relegation executions with granted authorization should succeed

Fix #615


tomtau (commented):
strangely some existing tests fail after these extra tests:
```
FAILED ../../../../../nix/store/wky88d6hfr01rcdac187a63pc9gfn5qa-source/integration_tests/test_basic.py::test_transfer
FAILED ../../../../../nix/store/wky88d6hfr01rcdac187a63pc9gfn5qa-source/integration_tests/test_basic.py::test_liquid_supply
```

Is it because they are executed on the same cluster?

MaCong-crypto:
> strangely some existing tests fail after these extra tests:
> 
> ```
> FAILED ../../../../../nix/store/wky88d6hfr01rcdac187a63pc9gfn5qa-source/integration_tests/test_basic.py::test_transfer
> FAILED ../../../../../nix/store/wky88d6hfr01rcdac187a63pc9gfn5qa-source/integration_tests/test_basic.py::test_liquid_supply
> ```
> 
> Is it because they are executed on the same cluster?

Yes, it is because they are executed on the same cluster.
I am working on some fixes to those test cases.

MaCong-crypto:
For these two failed test cases, `test_periodic_fee_allowance` and `test_exceed_period_limit_should_not_affect_the_next_period`. I will fix them in #626

integration_tests/test_authz.py
```diff
@@ -0,0 +1,382 @@
...
+    granter_address = cluster.address("community")
+    grantee_address = cluster.address("ecosystem")
+
+    grant_authorization(
+        cluster,
+        grantee_address,
+        AUTHORIZATION_DELEGATE,
+        granter_address,
+        spend_limit="%s%s" % (spend_limit, BASECRO_DENOM),
+        deny_validators=deny_validator_address,
```
tomtau, line 253: and is the logic that one can delegate to any validators except the ones on the deny list? not sure if the cluster in this test suite has multiple validators, but perhaps that could be good to check as well (one can't delegate to denied validators, but can delegate to others?)

tomtau (changes requested):
the tests are looking good -- great job!
added a few comments

integration_tests/utils.py
```diff
@@ -261,11 +284,181 @@ def revoke_fee_grant(cli, granter_address, grantee, i=0, *k_options, **kv_option
...
+            home=cli.cosmos_cli(i).data_dir,
+            keyring_backend="test",
+            chain_id=cli.cosmos_cli(i).chain_id,
+            node=cli.cosmos_cli(i).node_rpc,
+            **kv_options,
+        )
+    )
+
+
+def throw_error_for_non_success_code(func):
```
yihuang, line 297: nice

integration_tests/utils.py
```diff
@@ -261,11 +284,181 @@ def revoke_fee_grant(cli, granter_address, grantee, i=0, *k_options, **kv_option
...
+            src_validator,
+            dst_validator,
+            amount,
+            "-y",
+            *k_options,
+            from_=from_,
+            home=cli.cosmos_cli(i).data_dir,
+            keyring_backend="test",
+            chain_id=cli.cosmos_cli(i).chain_id,
+            node=cli.cosmos_cli(i).node_rpc,
```
yihuang, line 450:
```suggestion
```

I think we can neglect these parameters in cli in cosmos-sdk 0.43, because they are already set in the `client.toml`, set `home` alone will include all those options.

yihuang (approved):
LGTM, only minor code cleanup suggestions.

integration_tests/test_authz.py
```diff
@@ -0,0 +1,382 @@
...
+        "%s%s" % (transaction_coins, BASECRO_DENOM),
+        GENERATE_ONLY,
+    )
+    with open(generated_tx_txt, "w") as opened_file:
+        json.dump(generated_tx_msg, opened_file)
+
+    with pytest.raises(
+        Exception, match=r".*requested amount is more than spend limit.*"
+    ):
+        exec_tx_by_grantee(cluster, generated_tx_txt, grantee_address)
```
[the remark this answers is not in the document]
MaCong-crypto: Yeah, when I was writing the test, I added those asserts, but felt unnecessary and removed them.
Perhaps adding those asserts helps with readability and make the tests more self-explanatory , I will add them.

integration_tests/utils.py
MaCong-crypto: Yes, tried locally and verified that they can be removed. I will make the change.
Previously, I copied this code chunk from pystarport. Thank you for the advice 👍

integration_tests/test_authz.py
MaCong-crypto: One cannot delegate to denied validators and any other validators that are not in allowed validators. I think this `deny-validator` flag is just stating some denied validators explicitly.

MaCong-crypto:
Test failure will be fixed by #626

PR #632, opened 2021-08-25
Problem: latest IBC relayer release is not tested

tomtau:
Solution: bumped the hermes dependency to 0.7
in nix integration tests


devashishdxt (approved):

tomtau:
the failure is this https://github.com/crypto-org-chain/chain-main/issues/626#issuecomment-903655616

PR #638, opened 2021-08-30
Problem: test_periodic_fee_allowance occasionally fails on GH Actions (Fix #626)

### LINKED ISSUE #626: Problem: test_periodic_fee_allowance occasionally fails on GH Actions

MaCong-crypto:
Solution:

Wait for latest block time to pass transaction block time plus grant period, instead of sleeping for the grant period amount of time

Fix #626

### DIFF
--- a/integration_tests/test_feegrant.py
+++ b/integration_tests/test_feegrant.py
@@ -5,7 +5,6 @@ from dateutil.parser import isoparse
 
 from .utils import (
     BASECRO_DENOM,
-    BLOCK_BROADCASTING,
     SUCCESS_CODE,
     grant_fee_allowance,
     query_block_info,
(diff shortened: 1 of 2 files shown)


integration_tests/test_feegrant.py
```diff
@@ -153,15 +157,22 @@ def test_periodic_fee_allowance(cluster):
...
     for _ in range(number_of_periods):
-        transfer(
+        tx = transfer(
             cluster,
             fee_grantee_address,
             receiver_address,
             "%s%s" % (transaction_coins, BASECRO_DENOM),
             fees="%s%s" % (fee_coins, BASECRO_DENOM),
             fee_account=fee_granter_address,
+            broadcast_mode=BLOCK_BROADCASTING,
```
yihuang, line 167: I think `block` is the default option set in `client.toml`, no?

MaCong-crypto: Yeah, one of the `client.toml` looks like this
```
chain-id = "chaintest"
keyring-backend = "test"
output = "json"
node = "tcp://localhost:10017"
broadcast-mode = "block"
```
I will remove this line then.

yihuang (approved):

tomtau (approved):

Issue #618, opened 2021-08-11
Problem: potentially vulnerable dependencies in 0.43 upgrade


tomtau:
```
go list -json -m all | nancy sleuth

Vulnerable Packages

[1/1]	pkg:golang/github.com/coreos/etcd@3.3.13
3 known vulnerabilities affecting installed version
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ [CVE-2020-15114] In etcd before versions 3.3.23 and 3.4.10, the etcd gateway is a simple TCP prox...                                                                                                                                   ┃
┣━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┫
┃ Description        ┃ In etcd before versions 3.3.23 and 3.4.10, the etcd gateway is a simple TCP                                                                                                                                       ┃
┃                    ┃ proxy to allow for basic service discovery and access. However, it is                                                                                                                                             ┃
┃                    ┃ possible to include the gateway address as an endpoint. This results in a                                                                                                                                         ┃
┃                    ┃ denial of service, since the endpoint can become stuck in a loop of                                                                                                                                               ┃
┃                    ┃ requesting itself until there are no more available file descriptors to                                                                                                                                           ┃
┃                    ┃ accept connections on the gateway.                                                                                                                                                                                ┃
[...]

PR #643, opened 2021-09-02
Problem: latest v0.44 SDK is not used

tomtau:
Solution: bumped SDK + upgraded Go to 1.17

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,10 +1,10 @@
 module github.com/crypto-org-chain/chain-main/v3
 
-go 1.16
+go 1.17
 
 require (
 	github.com/confluentinc/bincover v0.1.0
-	github.com/cosmos/cosmos-sdk v0.43.0
+	github.com/cosmos/cosmos-sdk v0.44.0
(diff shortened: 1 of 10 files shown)


go.mod
```diff
@@ -26,6 +26,103 @@ require (
 
 )
 
+require (
```
tomtau, line 29: this is generated when I ran `go mod tidy` https://golang.org/doc/go1.17

tomtau: > For the go command to correctly resolve transitive imports using the pruned module graph, the go.mod file for each module needs to include more detail about the transitive dependencies relevant to that module. If a module specifies go 1.17 or higher in its go.mod file, its go.mod file now contains an explicit require directive for every module that provides a transitively-imported package. (In previous versions, the go.mod file typically only included explicit requirements for directly-imported packages.)

tomtau:
for the dependency stuff: https://github.com/crypto-org-chain/chain-main/issues/618

yihuang (approved):
LGTM

PR #646, opened 2021-09-02
Problem: testnet-croeseid-4 started on 0.43 

tomtau:
Solution: added a custom empty upgrade handler
to upgrade from 0.43 to 0.44


tomtau (commented):
@lezzokafka 
> Leslie Cheung  4 minutes ago
> Looking good to me. Just to confirm, by saying empty Upgrade Handler
when I submit the proposal, I should put
chain-maind tx gov submit-proposal software-upgrade croeseid-0.44 
as in here ?

PR #649, opened 2021-09-06
Problem: integration test data directory in CI is not accessible for debugging purpose (Fix #639)

### LINKED ISSUE #639: Problem: integration test data directory in CI is not accessible for debugging purpose
Currently, when integration test failed in CI, the only thing to debug is the aggregated logs in stdout. But I think we can upload the whole data directory as artifact. And with this we can keep stdout cleaner by not output all log files into stdout.

I did a test with this approach here: https://github.com/crypto-org-chain/cronos/actions/runs/1184425030

MaCong-crypto:
Solution:

Upload data directory as artifact upon test failure

Fix #639

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -61,6 +61,15 @@ jobs:
           docker login -u "$GITHUB_ACTOR" -p "${{ github.token }}" docker.pkg.github.com
           docker push $TARGET
           echo "pushed: $TARGET"
+      - name: Tar debug files
+        if: failure()
+        run: tar cfz debug_files.tar.gz -C /tmp/pytest-of-runner .
+      - uses: actions/upload-artifact@v2
+        if: failure()
+        with:
+          name: debug_files
+          path: debug_files.tar.gz
+          if-no-files-found: ignore
 
   test-upgrade:
     runs-on: ubuntu-latest
@@ -74,6 +83,15 @@ jobs:
           name: crypto-com
       - name: Run integration tests
         run: make nix-integration-test-upgrade
+      - name: Tar debug files
+        if: failure()
+        run: tar cfz debug_files_upgrade.tar.gz -C /tmp/pytest-of-runner .
+      - uses: actions/upload-artifact@v2
+        if: failure()
+        with:
+          name: debug_files_upgrade
+          path: debug_files_upgrade.tar.gz
+          if-no-files-found: ignore
 
   test-ledger:
     runs-on: ubuntu-latest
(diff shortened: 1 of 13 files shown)


tomtau (approved):

MaCong-crypto:
rebased on master and pushed again

yihuang:
I guess we could also change [the default value of supervisord-quiet flag](https://github.com/crypto-org-chain/chain-main/blob/master/integration_tests/conftest.py#L25) to `True` to keep the stdout cleaner?

yihuang (approved):
LGTM, on suggestion is to set [the default value of supervisord-quiet flag](https://github.com/crypto-org-chain/chain-main/blob/master/integration_tests/conftest.py#L25) to `True` to keep the stdout cleaner?

MaCong-crypto:
> LGTM, one suggestion is to set [the default value of supervisord-quiet flag](https://github.com/crypto-org-chain/chain-main/blob/master/integration_tests/conftest.py#L25) to `True` to keep the stdout cleaner?

I will try first and update here.

yihuang:
> > LGTM, one suggestion is to set [the default value of supervisord-quiet flag](https://github.com/crypto-org-chain/chain-main/blob/master/integration_tests/conftest.py#L25) to `True` to keep the stdout cleaner?
> 
> I will try first and update here.

Maybe just remove that flag all together, I think it's mainly added for CI debugging purpose.

tomtau:
will quiet supervisor still have some logs? one thing I worry is a "silent" issue when some integration tests keep failing, but due to some bug, it's reflected as pass in GH action output, so one way to notice it early would be from supervisor logs?

yihuang:
> will quiet supervisor still have some logs? one thing I worry is a "silent" issue when some integration tests keep failing, but due to some bug, it's reflected as pass in GH action output, so one way to notice it early would be from supervisor logs?

The quiet one doesn't tail the log files to the stdout, the normal stdout from pytest and test cases themselves are still displayed.

MaCong-crypto:
So far, I see all our automated test cases have been setting `supervisord-quiet` option to False, which means `tailer = cluster.start_tail_logs_thread(data)`  will be executed and `node.log` and `relayer.log` will be generated.
Unless when we run our test cases with `supervisord-quiet` to True explicitly, then we won't get those log files.

yihuang:
> So far, I see all our automated test cases have been setting `supervisord-quiet` option to False, which means `tailer = cluster.start_tail_logs_thread(data)` will be executed and `node.log` and `relayer.log` will be generated.
> Unless when we run our test cases with `supervisord-quiet` to True explicitly, then we won't get those log files.

The log files are still there, just not displayed in stdout.

MaCong-crypto:
> > So far, I see all our automated test cases have been setting `supervisord-quiet` option to False, which means `tailer = cluster.start_tail_logs_thread(data)` will be executed and `node.log` and `relayer.log` will be generated.
> > Unless when we run our test cases with `supervisord-quiet` to True explicitly, then we won't get those log files.
> 
> The log files are still there, just not displayed in stdout.

I tried just now, log files are there, just not displayed in stdout as you described. I misunderstood this function `start_tail_logs_thread()`. 👍

PR #657, opened 2021-09-30
Problem: upstream SDK release not used

tomtau:
Solution: changed the version from a custom fork to the upstream release

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.16
 
 require (
 	github.com/confluentinc/bincover v0.1.0
-	github.com/cosmos/cosmos-sdk v0.44.0
+	github.com/cosmos/cosmos-sdk v0.44.1
 	github.com/cosmos/ibc-go v1.2.0
 	github.com/cosmos/ledger-go v0.9.2 // indirect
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -38,5 +38,4 @@ replace github.com/99designs/keyring => github.com/crypto-org-chain/keyring v1.1
 
 replace github.com/dgrijalva/jwt-go => github.com/dgrijalva/jwt-go/v4 v4.0.0-preview1
 
-// FIXME: back to upstream release once upgrade non-determinism is gone
-replace github.com/cosmos/cosmos-sdk => github.com/crypto-org-chain/cosmos-sdk v0.44.0-upgradefix
+replace github.com/cosmos/cosmos-sdk => github.com/cosmos/cosmos-sdk v0.44.1
```
yihuang, line 41: is the replace still needed?

tomtau: ibc-go 1.2.0 depends on 0.44.0 -- without the replace, 0.44.0 will still be in go.sum... so just put it there to force 0.44.1

yihuang (approved):

PR #674, opened 2021-11-04
Problem: Missing ADR draft for NFT transition to the SDK module

devashishdxt:
Solution: Added a draft ADR for transition

### DIFF
--- a/doc/architecture/adr-004.md
+++ b/doc/architecture/adr-004.md
@@ -8,7 +8,7 @@
 
 Currently, Crypto.org Chain offers support for non-fungible tokens (NFTs) using a custom NFT module in `chain-main`. The
 current implementation of NFT module is based on [Irismod's NFT module](https://github.com/irisnet/irismod/tree/master/modules/nft).
-Given that different chain's are using their custom implementation of NFT module, it is very difficult to achieve
+Given that different chains are using their custom implementation of NFT module, it is very difficult to achieve
 flawless interportability of NFTs across multiple chains using IBC. At the same time, there are more usecases that needs
 to be supported using NFTs which may require IBC support. So, there is a need of a standardized NFT implementation
 which has support for reusing NFTs between modules and chains.
--- a/doc/architecture/adr-004.md
+++ b/doc/architecture/adr-004.md
@@ -254,12 +254,13 @@ for free (for example, IBC support, inter-module asset support using [ADR-33](ht
 
 ### Negative
 
-Our current implementation is slightly different from Cosmos SDK's NFT module which can result in backwards
-incompatibility. This can be temporarily avoided by creating a wrapper module (which we anyway need to create for 
-minting and burning NFTs) and adding same messages as the current module along with the sandard message types mentioned
-[here](https://github.com/cosmos/cosmos-sdk/blob/master/docs/architecture/adr-043-nft-module.md#abstract). But, it is
-necessary to have a deprecation plan for old message types and transition all the downstream services to new
-standardized message types.
+- Our current implementation is slightly different from Cosmos SDK's NFT module which can result in backwards
+  incompatibility. This can be temporarily avoided by creating a wrapper module (which we anyway need to create for 
+  minting and burning NFTs) and adding same messages as the current module along with the sandard message types mentioned
+  [here](https://github.com/cosmos/cosmos-sdk/blob/master/docs/architecture/adr-043-nft-module.md#abstract). But, it is
+  necessary to have a deprecation plan for old message types and transition all the downstream services to new
+  standardized message types.
+- There'll be a lot of duplicate message types until legacy messages are removed.
 
 ### Neutral
 
--- a/doc/architecture/adr-004.md
+++ b/doc/architecture/adr-004.md
@@ -6,7 +6,7 @@
 
 ## Context
 
-Currently, Crypto.com Chain offers support for non-fungible tokens (NFTs) using a custom NFT module in `chain-main`. The
+Currently, Crypto.org Chain offers support for non-fungible tokens (NFTs) using a custom NFT module in `chain-main`. The
 current implementation of NFT module is based on [Irismod's NFT module](https://github.com/irisnet/irismod/tree/master/modules/nft).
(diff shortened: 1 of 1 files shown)


doc/architecture/adr-004.md
```diff
@@ -0,0 +1,82 @@
...
+
+   **Advantages**
+
+   - It'll be very easy to add custom logic base on our requirements (for example, https://github.com/irisnet/irismod/issues/183).
+
+   **Drawbacks**
+
+   - Because this module will be developed in isolation, we'll have to work extra hard to make sure that it confirms to
+     Cosmos SDK's NFT standards.
+   - Custom logic added to the module may make it impossible to stay in sync with Cosmos SDK's standards.
```
tomtau, line 50: given this is not a decision to take, it can be removed from here

doc/architecture/adr-004.md
```diff
@@ -0,0 +1,82 @@
...
+## Consequences
+
+### Positive
+
+Cosmos SDK's module will have support from multiple members of the community and will get all the future improvements
+for free (for example, IBC support, inter-module asset support using [ADR-33](https://github.com/cosmos/cosmos-sdk/blob/master/docs/architecture/adr-033-protobuf-inter-module-comm.md))
+
+### Negative
+
+Our current implementation is slightly different from Cosmos SDK's NFT module which can result in backwards
```
tomtau, line 67: > Our current implementation is slightly different from Cosmos SDK's NFT module which can result in backwards
 incompatibility.

it would be good to summarise the details of what exactly is different

>  But, it is necessary to have a deprecation plan for old message types and transition all the downstream services to new standardized message types.

the plan can be sketched out in the decision section

tomtau (changes requested):
it's a good start! If at this stage, we already know the module differences and anticipate how the transition will go, it'll be good to add more details:
- mention the differences
- sketch out the state transition code for the upgrade handler 
- all things that should happen (bump the module version? gRPC and message routing? any metadata changes? ...)

tomtau:
For the wrapper module: is that the existing NFT module can be changed to it?
e.g.`MsgNewClass` would be the same as `MsgIssueDenom`? (so both message types would exist, but the message handler logic would be just one?)
So one more consequence (neutral or negative) is that there'll be duplicate message types?

IngridMi (approved):

doc/architecture/adr-004.md
devashishdxt: Done

doc/architecture/adr-004.md
devashishdxt: I wanted to keep it here so that in future, we know all the things that we considered.

devashishdxt:
> For the wrapper module: is that the existing NFT module can be changed to it?
> e.g.`MsgNewClass` would be the same as `MsgIssueDenom`? (so both message types would exist, but the message handler logic would be just one?)
> So one more consequence (neutral or negative) is that there'll be duplicate message types?

Done

doc/architecture/adr-004.md
```diff
@@ -0,0 +1,273 @@
+# Transition to Cosmos SDK's NFT module
+
+## Changelog
+
+- 04-11-2021: Initial Draft
+
+## Context
+
+Currently, Crypto.com Chain offers support for non-fungible tokens (NFTs) using a custom NFT module in `chain-main`. The
```
tomtau, line 9:
```suggestion
Currently, Crypto.org Chain offers support for non-fungible tokens (NFTs) using a custom NFT module in `chain-main`. The
```

doc/architecture/adr-004.md
```diff
@@ -0,0 +1,273 @@
...
+current implementation of NFT module is based on [Irismod's NFT module](https://github.com/irisnet/irismod/tree/master/modules/nft).
+Given that different chain's are using their custom implementation of NFT module, it is very difficult to achieve
```
tomtau, line 11:
```suggestion
Given that different chains are using their custom implementation of NFT module, it is very difficult to achieve
```

doc/architecture/adr-004.md
```diff
@@ -0,0 +1,273 @@
...
+flawless interportability of NFTs across multiple chains using IBC. At the same time, there are more usecases that needs
```
tomtau, line 12:
```suggestion
flawless interportability of NFTs across multiple chains using IBC. At the same time, there are more usecases that need
```

doc/architecture/adr-004.md
```diff
@@ -0,0 +1,273 @@
...
+
+### Negative
+
+- Our current implementation is slightly different from Cosmos SDK's NFT module which can result in backwards
+  incompatibility. This can be temporarily avoided by creating a wrapper module (which we anyway need to create for 
+  minting and burning NFTs) and adding same messages as the current module along with the sandard message types mentioned
+  [here](https://github.com/cosmos/cosmos-sdk/blob/master/docs/architecture/adr-043-nft-module.md#abstract). But, it is
+  necessary to have a deprecation plan for old message types and transition all the downstream services to new
+  standardized message types.
+- There'll be a lot of duplicate message types until legacy messages are removed.
```
tomtau, line 263: the only duplicate message type should be transfer? the others are needed as the wrapper?

tomtau (approved):
looks good for the draft! 👍 there are a few missing details (e.g. deprecation plan, gRPC -- I guess we'll also keep both? -- or the storage transition), but we can sort these out out later.
@kaantig @chrisyang-crypto any final comment?

cdc-Hitesh (approved):

kaantig:
> looks good for the draft! 👍 there are a few missing details (e.g. deprecation plan, gRPC -- I guess we'll also keep both? -- or the storage transition), but we can sort out these out later. @kaantig @chrisyang-crypto any final comment?

LGTM!

doc/architecture/adr-004.md
devashishdxt: No. There'll be duplicate messages even for other things (minting, burning, etc.). One will be the current messages and other will be the messages specified in Cosmos SDK's ADR.

PR #688, opened 2021-12-13
Problem: no rapid setup of performant relay nodes

faddat:
# Solution

This PR adds scripts/statesync.bash, which will have a relay-optoimized state synced node running in seconds.  The goal of this commit is to allow relayers to run with the setup that I find ideal:

* State Synced
* RocksDB
* Latest Tendermint

In on order for rocksdb to work correctly, it replaces tecbot/gorocksdb with cosmos/gorocksdb.

It bumps the Tendermint version because .15 seems to do a bit better than .14.

# PR Checklist:

- [x] Have you read the [CONTRIBUTING.md](https://github.com/crypto-org-chain/chain-main/blob/master/CONTRIBUTING.md)?
- [x] Have you rebased your work on top of the latest master? 
- [x] Have you checked your code compiles? (`make`)
- [x] Have you included tests for any non-trivial functionality?
- [x] If you are contributing for the first time, please read the agreement in [CONTRIBUTING.md](https://github.com/crypto-org-chain/chain-main/blob/master/CONTRIBUTING.md) now and add a comment to this pull request stating that your PR is in accordance with the [Developer's Certificate of Origin](https://github.com/crypto-org-chain/chain-main/blob/master/CONTRIBUTING.md#developer-certificate-of-originn).

Yes, I agree with contributing.md and will shortly get a little more info on DCO.

Yes, I'm the author of this code :). 

Yes, I like outlining PR's as problem and solution.  Have a nice day!

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,8 +17,8 @@ require (
 	github.com/spf13/cobra v1.2.1
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.7.0
-	github.com/tendermint/tendermint v0.34.14
-	github.com/tendermint/tm-db v0.6.6
+	github.com/tendermint/tendermint v0.34.15
+	github.com/tendermint/tm-db v0.6.7-0.20211203155021-4fa83b55a0b5
 	github.com/tidwall/gjson v1.9.3
 	google.golang.org/genproto v0.0.0-20210917145530-b395a37504d4
 	google.golang.org/grpc v1.42.0
@@ -144,5 +144,7 @@ replace github.com/opencontainers/image-spec => github.com/opencontainers/image-
 replace github.com/cosmos/cosmos-sdk => github.com/cosmos/cosmos-sdk v0.44.5
 
 replace github.com/tendermint/tendermint => github.com/tendermint/tendermint v0.34.15
+
+replace github.com/tecbot/gorocksdb => github.com/cosmos/gorocksdb v0.0.0-20211202124722-2c356d6d98e4
 
 replace github.com/cosmos/iavl => github.com/cosmos/iavl v0.17.3
--- a/scripts/statesync.bash
+++ b/scripts/statesync.bash
@@ -1,0 +1,41 @@
+#!/bin/bash
+# microtick and bitcanna contributed significantly here.
+set -uxe
+
+# set environment variables
+export GOPATH=~/go
+export PATH=$PATH:~/go/bin
+
+
+# Install chain-maind
+go install -tags rocksdb ./...
+
+# MAKE HOME FOLDER AND GET GENESIS
+chain-maind init test --home /gaia2/cro 
+wget -O /gaia2/cro/config/genesis.json https://github.com/crypto-org-chain/mainnet/raw/main/crypto-org-chain-mainnet-1/genesis.json
+
+INTERVAL=1000
+
+# GET TRUST HASH AND TRUST HEIGHT
+
+LATEST_HEIGHT=$(curl -s https://mainnet.crypto.org/block | jq -r .result.block.header.height);
+BLOCK_HEIGHT=$(($LATEST_HEIGHT-$INTERVAL)) 
+TRUST_HASH=$(curl -s "https://mainnet.crypto.org/block?height=$BLOCK_HEIGHT" | jq -r .result.block_id.hash)
+
+
+# TELL USER WHAT WE ARE DOING
+echo "TRUST HEIGHT: $BLOCK_HEIGHT"
+echo "TRUST HASH: $TRUST_HASH"
+
+
+# export state sync vars
+export CHAIN_MAIND_GRPC_WEB_ADDRESS=127.0.0.1:9876
+export CHAIN_MAIND_STATESYNC_ENABLE=true
+export CHAIN_MAIND_P2P_MAX_NUM_OUTBOUND_PEERS=200
+export CHAIN_MAIND_STATESYNC_RPC_SERVERS="https://mainnet.crypto.org:443,https://mainnet.crypto.org:443"
+export CHAIN_MAIND_STATESYNC_TRUST_HEIGHT=$BLOCK_HEIGHT
+export CHAIN_MAIND_STATESYNC_TRUST_HASH=$TRUST_HASH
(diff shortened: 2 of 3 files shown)


tomtau (commented):
thanks for the contribution! If I understand it correctly, the problem this is trying to solve is that the default build setup isn't ideal for high performance sentry nodes, is that correct? Could you perhaps open an issue that would have a more detailed problem description and link it here?

Once that's clearer, one thing to consider: why RocksDB and not BadgerDB? (BadgerDB should be supported too and it should be easier to build, given it's pure Go.)

Plus, if the problem is about the default build config, the solution for an alternative build script may be missing e.g. ldflags ( `-X github.com/cosmos/cosmos-sdk/types.DBBackend=...`): https://github.com/cosmos/cosmos-sdk/blob/52f6fc29da96339787da2048bb326b747a71c798/Makefile#L64

faddat:
@tomtau 

Basically, relay nodes across Cosmos -- Osmosis, Gaia, Terra and Cro in particular-- are hit with an unusual isssue.  Nodes used for relaying resemble sentries except that most sentries do not have very much inbound query traffic / tx writes.  Relay nodes have lots of both.

When the IBC network heats up, the node software is no longer able to sync and keep up to the tip of the chain. 

I imagine that you noticed that as the CRO chain gets longer, it gets less and less fast, while consuming more and more disk i/o bandwith, right?

Well, state sync is a way to remedy this issue because you've got only the tip of the chain.  Additionally, state sync makes it possible to -- no kidding -- validate gaia from a raspberry pi.  That's how much disk i/o consumption is reduced (I think by "shortening" the iavl tree, so there's less balancing to do.  So, when you've got a state synced node, another benefit is that it does far better in terms of query performance and writes.  

Typically, I'm in full support of 100% go solutions, but the reason for the ibc traffic jams is basically that relayers like us at Notional or the CROsNest team aren't able to keep nodes synced to the tip of the chain.  So far, my testing has shown that badgerdb is just a little bit faster than goleveldb for use on tendermint chains, while RocksDB is **much** faster, so this particular script is actually part of a devops deployment pattern we're building out at Notional to make our relay setup reproducible ("just add seed phrase") and much more performant than it currently is.  

Long-run, the team at valnodes has proposed:

https://twitter.com/valnodes/status/1466903567013449732?s=20

And I'm in full support of that effort but I figure it is at least a year out currently.

tomtau (changes requested):
ok, could you add the missing parts for the rocksdb build: https://github.com/cosmos/cosmos-sdk/blob/52f6fc29da96339787da2048bb326b747a71c798/Makefile#L76 in the script?

scripts/statesync.bash
```diff
@@ -0,0 +1,41 @@
...
+
+# export state sync vars
+export CHAIN_MAIND_GRPC_WEB_ADDRESS=127.0.0.1:9876
+export CHAIN_MAIND_STATESYNC_ENABLE=true
+export CHAIN_MAIND_P2P_MAX_NUM_OUTBOUND_PEERS=200
+export CHAIN_MAIND_STATESYNC_RPC_SERVERS="https://mainnet.crypto.org:443,https://mainnet.crypto.org:443"
+export CHAIN_MAIND_STATESYNC_TRUST_HEIGHT=$BLOCK_HEIGHT
+export CHAIN_MAIND_STATESYNC_TRUST_HASH=$TRUST_HASH
+export CHAIN_MAIND_P2P_SEEDS="8dc1863d1d23cf9ad7cbea215c19bcbe8bf39702@p2p.baaa7e56-cc71-4ae4-b4b3-c6a9d4a9596a.cryptodotorg.bison.run:26656,8a7922f3fb3fb4cfe8cb57281b9d159ca7fd29c6@p2p.aef59b2a-d77e-4922-817a-d1eea614aef4.cryptodotorg.bison.run:26656,494d860a2869b90c458b07d4da890539272785c9@p2p.fabc23d9-e0a1-4ced-8cd7-eb3efd6d9ef3.cryptodotorg.bison.run:26656,dc2540dabadb8302da988c95a3c872191061aed2@p2p.7d1b53c0-b86b-44c8-8c02-e3b0e88a4bf7.cryptodotorg.herd.run:26656,"
+export CHAIN_MAIND_P2P_PERSISTENT_PEERS="87c3adb7d8f649c51eebe0d3335d8f9e28c362f2@seed-0.crypto.org:26656,e1d7ff02b78044795371beb1cd5fb803f9389256@seed-1.crypto.org:26656"
```
yihuang, line 39: nice, I forgot we can override configs in this way.

faddat:
Happy to add the rocksdb config to the Makefile :).

tomtau:
there's a small conflict with the latest master -- @faddat could you rebase your PR after adding `-X github.com/cosmos/cosmos-sdk/types.DBBackend=rocksdb` or what else is needed in the build script?

go.mod
```diff
@@ -145,4 +145,6 @@ replace github.com/cosmos/cosmos-sdk => github.com/cosmos/cosmos-sdk v0.44.5
 
 replace github.com/tendermint/tendermint => github.com/tendermint/tendermint v0.34.15
 
+replace github.com/tecbot/gorocksdb => github.com/cosmos/gorocksdb v0.0.0-20211202124722-2c356d6d98e4
```
yihuang, line 148: https://github.com/cosmos/cosmos-sdk/blob/master/db/go.mod#L36
cosmos-sdk use `v1.1.1`, should we follow suite here?

scripts/statesync.bash
```diff
@@ -0,0 +1,41 @@
...
+# microtick and bitcanna contributed significantly here.
+set -uxe
+
+# set environment variables
+export GOPATH=~/go
+export PATH=$PATH:~/go/bin
+
+
+# Install chain-maind
+go install -tags rocksdb ./...
```
yihuang, line 11: One need to install the rocksdb c library by himself in advance, right?

yihuang: Do we also need the ldflat: `-X github.com/cosmos/cosmos-sdk/types.DBBackend=rocksdb`?

tomtau:
Closing this PR for now -- feel free to reopen it with the extra changes in Makefile

PR #690, opened 2021-12-16
Problem: incorrect module name in OnChanOpenInit events

tomtau:
Solution: bumped the ibc-go dependency to 2.0.2


ysong42 (approved):
LGTM! It will not affect the chain-indexing.

PR #724, opened 2022-03-10
Problem: `x/supply` is not marked as deprecated

### LINKED ISSUE #720: Problem: `x/supply` is not marked as deprecated
As discussed in #716, we can deprecate `x/supply` module and eventually remove it when an alternate way to query liquid supply is available (either via cosmos-sdk's native module or chain-indexer).

Ref: https://github.com/protocolbuffers/protobuf/issues/1734#issuecomment-662623725

- [x] Write an ADR (#723)
- [ ] Implement changes (#724)

devashishdxt:
Solution: Marked `x/supply` grpc and cli as deprecated. Fixes #720.


x/nft/types/query.pb.gw.go
```diff
@@ -20,7 +20,6 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/grpclog"
-	"google.golang.org/grpc/metadata"
```
tomtau, line 23: what are these changes in the generated pb?

tomtau (approved):

x/nft/types/query.pb.gw.go
devashishdxt: Not sure. Should I revert these? Integration tests seems to work fine with these changes.

devashishdxt: Maybe because of different `protoc` version on my machine?

tomtau: I'm not sure whether these metadata settings are anyhow critical for e.g. security headers or if one puts it behind LB. @yihuang do you know?

 maybe the best is to align it with how the generated code is in other projects? (cronos, ethermint, Cosmos SDK...)

devashishdxt: I've confirmed in Cosmos SDK. `.pb.gw.go` files in Cosmos SDK also doesn't have import for `google.golang.org/grpc/metadata` and any function call to `NewContextWithServerTransportStream`

devashishdxt: Merging this PR.

Issue #731, opened 2022-03-17
Problem: Undecided default `TimeoutDuration` for interchain account packets


devashishdxt:
Currently, the default `TimeoutDuration` for interchain account packets (when sending a transaction to host chain on behalf of interchain account) is 5 minutes (which can be overridden by the caller). Since interchain accounts use ordered channels, the channel will close if any packet times out. 5 minutes is a safe duration for relaying packets.

There are four options:

1. Keep default timeout to 5 minutes
2. Change default timeout to some other value
3. Remove the default timeout and make it mandatory for the caller to provide it in request.
4. Add `DefaultTimeoutDuration` to `icaauth` params so that it can be configured and changed in future.

@tomtau @yihuang @calvinaco Please provide your inputs in comments.

tomtau:
@allthatjazzleo what was the longest delay for acking ics20?

tomtau:
@devashishdxt should the value also be meant as a minimum? if someone puts a smaller timeout that could happen, it'll close a channel?

devashishdxt:
> @devashishdxt should the value also be meant as a minimum? if someone puts a smaller timeout that could happen, it'll close a channel?

Yes. If someone puts a smaller timeout, then the channel may close if the packet times out. I think it is a good idea to make this as a `MinimumTimeoutDuration`. Should we return an error when the caller provided timeout duration is smaller than this? Also, if the caller doesn't provide any timeout duration, should this be treated as default or an error should be returned?

tomtau:
how is it done in other ibc modules? maybe good to keep it consistent?

devashishdxt:
> how is it done in other ibc modules? maybe good to keep it consistent?

Current default for `transfer` module in `ibc-go` is `10 minutes`. https://github.com/cosmos/ibc-go/blob/dbd2df26e54f17f94a498d3fe048d68df04735b6/modules/apps/transfer/types/packet.go#L21

`transfer` module completely disables the timeout if it is set to zero. But, I don't think that is possible for us to do here because `ica` module uses ordered channels.

tomtau:
and other ibc apps? not sure if there's one more that also uses ordered channels?

devashishdxt:
> and other ibc apps? not sure if there's one more that also uses ordered channels?

I'm not aware of any IBC apps other than transfer and interchain accounts.

calvinaco:
Because of the nature of ordered channel, summarizing the above comment, how about:

[...]

Issue #726, opened 2022-03-11
Problem: Undecided `AllowMessages` for interchain accounts 


devashishdxt:
Interchain accounts module has an ability to specify the message type or transactions that interchain accounts are authorised to execute. This issue is just for discussion around it and reaching a conclusion on which message types should be enabled for interchain accounts on `chain-main`

Ref: https://github.com/cosmos/ibc-go/blob/main/docs/apps/interchain-accounts/parameters.md#allowmessages

@tomtau @yihuang @calvinaco 

Please add your recommendations in the comments.

For me, `/cosmos.bank.v1beta1.MsgSend` is a no-brainer. But, do we also want to allow staking or governance related messages?

tomtau:
@devashishdxt I think staking will be important for liquid staking-related use cases + possibly for that "key rotation" use case via solo machines.

gov should be there as well -- or if not gov, then authz

devashishdxt:
List of `AllowMessages` in cosmos hub: https://github.com/cosmos/gaia/blob/f71a4be8b43c9891a36edbd365a5f3e8403da769/app/app.go#L639

devashishdxt:
Current PR to enable interchain accounts uses list similar to cosmos hub except for tendermint related messages: https://github.com/crypto-org-chain/chain-main/pull/732

calvinaco:
On a second thought, I think we can also add the authz messages too. This could potentially help with some advance liquid staking.

calvinaco:
On a second thought, I think we can also add the authz messages too. This could potentially help with some advance liquid staking.

PR #732, opened 2022-03-17
Problem: Interchain accounts are not enabled in `chain-main`

devashishdxt:
Solution: Enable interchain accounts in `app.go` and add an upgrade handler

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -534,28 +534,21 @@ func New(
 		hostParams := icahosttypes.Params{
 			HostEnabled: false,
 			AllowMessages: []string{
-				"/cosmos.authz.v1beta1.MsgExec",
-				"/cosmos.authz.v1beta1.MsgGrant",
-				"/cosmos.authz.v1beta1.MsgRevoke",
 				"/cosmos.bank.v1beta1.MsgSend",
 				"/cosmos.bank.v1beta1.MsgMultiSend",
 				"/cosmos.distribution.v1beta1.MsgSetWithdrawAddress",
 				"/cosmos.distribution.v1beta1.MsgWithdrawValidatorCommission",
 				"/cosmos.distribution.v1beta1.MsgFundCommunityPool",
 				"/cosmos.distribution.v1beta1.MsgWithdrawDelegatorReward",
-				"/cosmos.feegrant.v1beta1.MsgGrantAllowance",
-				"/cosmos.feegrant.v1beta1.MsgRevokeAllowance",
 				"/cosmos.gov.v1beta1.MsgVoteWeighted",
 				"/cosmos.gov.v1beta1.MsgSubmitProposal",
 				"/cosmos.gov.v1beta1.MsgDeposit",
 				"/cosmos.gov.v1beta1.MsgVote",
-				"/cosmos.staking.v1beta1.MsgEditValidator",
 				"/cosmos.staking.v1beta1.MsgDelegate",
 				"/cosmos.staking.v1beta1.MsgUndelegate",
 				"/cosmos.staking.v1beta1.MsgBeginRedelegate",
-				"/cosmos.staking.v1beta1.MsgCreateValidator",
-				"/cosmos.vesting.v1beta1.MsgCreateVestingAccount",
 				"/ibc.applications.transfer.v1.MsgTransfer",
+				"/chainmain.nft.v1.MsgTransferNFT",
 			},
 		}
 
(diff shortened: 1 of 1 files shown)


app/app.go
```diff
@@ -467,6 +522,48 @@ func New(
...
+		fromVM[icatypes.ModuleName] = icaModule.ConsensusVersion()
+
+		// create ICS27 Controller submodule params
+		controllerParams := icacontrollertypes.Params{
+			ControllerEnabled: true,
+		}
+
+		// create ICS27 Host submodule params
+		hostParams := icahosttypes.Params{
+			HostEnabled: true,
```
tomtau, line 535: maybe disable first in the upgrade handler and enable via a gov proposal later (as done with ICS20)?

devashishdxt: Changed both `ControllerEnabled` and `HostEnabled` to `false`.

app/app.go
```diff
@@ -467,6 +522,48 @@ func New(
...
+			ControllerEnabled: false,
+		}
+
+		// create ICS27 Host submodule params
+		hostParams := icahosttypes.Params{
+			HostEnabled: false,
+			AllowMessages: []string{
+				"/cosmos.authz.v1beta1.MsgExec",
+				"/cosmos.authz.v1beta1.MsgGrant",
+				"/cosmos.authz.v1beta1.MsgRevoke",
```
tomtau, line 539: maybe remove authz and feegrant "/cosmos.staking.v1beta1.MsgCreateValidator", "/cosmos.vesting.v1beta1.MsgCreateVestingAccount", and "/cosmos.staking.v1beta1.MsgEditValidator".... 
they seem a bit esoteric for interchain accounts?

and maybe add Nft transfer? it could be potentially interesting e.g. if a Nft is owned by an interchain account that corresponds to a smart contract on Cronos?

app/app.go
```diff
@@ -478,7 +575,9 @@ func New(
 	}
 
 	if upgradeInfo.Name == planName && !app.UpgradeKeeper.IsSkipHeight(upgradeInfo.Height) {
-		storeUpgrades := storetypes.StoreUpgrades{}
+		storeUpgrades := storetypes.StoreUpgrades{
+			Added: []string{icacontrollertypes.StoreKey, icahosttypes.StoreKey},
```
tomtau, line 579: the ica auth module doesn't need to be added?

tomtau (commented):
the app wiring seems all right

app/app.go
```diff
@@ -467,6 +522,48 @@ func New(
 
 	planName := "v4.0.0"
 	app.UpgradeKeeper.SetUpgradeHandler(planName, func(ctx sdk.Context, _ upgradetypes.Plan, fromVM module.VersionMap) (module.VersionMap, error) {
+		// set the ICS27 consensus version so InitGenesis is not run
+		fromVM[icatypes.ModuleName] = icaModule.ConsensusVersion()
```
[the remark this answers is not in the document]
devashishdxt: Because we're calling `InitModule` later in upgrade handler. 

Ref:

- https://github.com/cosmos/ibc-go/blob/46e020640e66f9043c14c53a4d215a5b457d6703/modules/apps/27-interchain-accounts/module.go#L107
- https://github.com/cosmos/ibc-go/blob/v3.0.0/docs/migrations/v2-to-v3.md#upgrade-proposal

app/app.go
devashishdxt: ica auth module does not access the store. Anyway, I've added it to the store upgrades just to be safe.

app/app.go
devashishdxt: Done

app/app.go
tomtau: and it doesn't have any params etc., so no need to initialize anything in the upgrade handler?
or should that 5minute timeout be a param?

devashishdxt: It doesn't have any params right now. But, it's a good idea to move default timeout duration to params. I'll do that in future PR.

devashishdxt: Added an option here: https://github.com/crypto-org-chain/chain-main/issues/731#issue-1171885779

tomtau (approved):
@calvinaco mentioned it may also be good to add minting/burning NFTs or perhaps denom issue?

devashishdxt:
> @calvinaco mentioned it may also be good to add minting/burning NFTs or perhaps denom issue?

Ok. I'll keep #726 open until we've reached a conclusion on this.

PR #734, opened 2022-03-21
Problem: `.golanci.yml` contains a lot of unncessary linters

devashishdxt:
Solution: Removed linters that are enabled by default and replaced deprecated linters with recommended ones + format code and fix lint errors

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,43 +1,21 @@
 linters:
   enable:
     - bodyclose
-    - deadcode
     - depguard
     - dogsled
     - dupl
-    - errcheck
     - exportloopref
-    # - funlen
-    # - gochecknoglobals
-    # - gochecknoinits
     - goconst
     - gocritic
-    # - gocyclo
-    # - godox
     - gofmt
     - goimports
-    - golint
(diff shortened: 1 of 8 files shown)


devashishdxt:
Looks like CI has an old version of `golangci-lint`

tomtau (commented):
maybe this value needs to be bumped: https://github.com/crypto-org-chain/chain-main/blob/master/.github/workflows/lint.yml#L26 ?

tomtau:
https://github.com/golangci/golangci-lint-action#compatibility

.golangci.yml
```diff
@@ -1,43 +1,21 @@
 linters:
   enable:
     - bodyclose
-    - deadcode
```
tomtau, line 4: why are those "unnecessary"? they seem like good defaults from Cosmos SDK? or is it they are covered by a different linter now

devashishdxt: These are enabled by default. No need to specify them. https://golangci-lint.run/usage/linters/

tomtau (approved):

PR #738, opened 2022-03-23
Problem: No way to change default packet timeout for `icaauth` module

### LINKED ISSUE #731: Problem: Undecided default `TimeoutDuration` for interchain account packets

devashishdxt:
Solution: Added `MinTimeoutDuration` to `icaauth` module parameters. Fixes #731.

- Also changed the protobuf generation code (it is now completely in-sync with cosmos-sdk)


app/app.go
```diff
@@ -755,6 +756,7 @@ func initParamsKeeper(appCodec codec.BinaryCodec, legacyAmino *codec.LegacyAmino
 	paramsKeeper.Subspace(ibchost.ModuleName)
 	paramsKeeper.Subspace(icacontrollertypes.SubModuleName)
 	paramsKeeper.Subspace(icahosttypes.SubModuleName)
+	paramsKeeper.Subspace(icaauthmoduletypes.ModuleName)
```
tomtau, line 760: I guess the upgrade handler should set this param to something?

app/docs/swagger-ui/swagger.yaml
```diff
@@ -4620,6 +4620,469 @@ paths:
...
+
+                        {
+                          "@type": "type.googleapis.com/google.profile.Person",
+                          "firstName": <string>,
+                          "lastName": <string>
+                        }
+
+                    If the embedded message type is well-known and has a custom
+                    JSON
+
```
tomtau, line 5012: maybe swagger.yaml can be patched in a separate PR?

proto/icaauth/v1/query.proto
```diff
@@ -4,21 +4,36 @@ package chainmain.icaauth.v1;
...
 import "cosmos/base/query/v1beta1/pagination.proto";
+import "icaauth/v1/params.proto";
 
 option go_package = "github.com/crypto-org-chain/chain-main/x/icaauth/types";
 
 // Query defines the gRPC querier service.
 service Query {
-    // InterchainAccountAddress queries the interchain account address for given `connectionId` and `owner`
-    rpc InterchainAccountAddress(QueryInterchainAccountAddressRequest) returns (QueryInterchainAccountAddressResponse) {
-        option (google.api.http).get = "/devashishdxt/mars/icaauth/interchain_account_address/{connectionId}/{owner}";
```
tomtau, line 14: oops, this one slipped the previous review

x/supply/types/query.pb.go
```diff
@@ -6,6 +6,7 @@ package types
 import (
 	context "context"
 	fmt "fmt"
+	github_com_cosmos_cosmos_sdk_types "github.com/cosmos/cosmos-sdk/types"
 	types "github.com/cosmos/cosmos-sdk/types"
```
tomtau, line 10: this is a bit strange that the types are duplicated? and why is the supply query updated?

tomtau (commented):
what is the "swagger.yaml.bak" file for?

app/docs/swagger-ui/swagger.yaml
tomtau: or maybe ok in this PR

app/app.go
devashishdxt: There are two ways a doing this AFAIK.

- Disable `InitGenesis` of a module and set params manually.
- Let the default params set by `InitGenesis`.

For interchain accounts, we are manually setting params (and disabling `InitGenesis`) because we want to modify the default params. But for `icaauth` we don't want to modify the default value of genesis params.

devashishdxt: I've added a test for this in `test_upgrade.py` by querying params after the upgrade.

app/docs/swagger-ui/swagger.yaml
devashishdxt: If it is difficult for you to review this, I can split it in two PRs?

x/supply/types/query.pb.go
devashishdxt: Because this PR changes the protobuf code generation logic (it is now exactly same as `cosmos-sdk`).

devashishdxt:
> what is the "swagger.yaml.bak" file for?

Removed

app/docs/swagger-ui/swagger.yaml
tomtau: it should be ok to review in this PR, but can do it in more depth tomorrow

app/app.go
tomtau: ok

tomtau (approved):

PR #741, opened 2022-03-24
Problem: Simulation tests fail after enabling ICA module

devashishdxt:
Solution: Vendored in `state.go` in `chain-main/app` so that it takes default genesis value from `app.ModuleBasic` instead of taking it from cosmos-sdk's simapp


app/state.go
```diff
@@ -0,0 +1,247 @@
+package app
```
tomtau, line 1: it'd be good to include a copyright header with licensing information if this is copied from Cosmos SDK

tomtau (changes requested):

app/state.go
devashishdxt: Done

yihuang (approved):
LGTM

tomtau (approved):

PR #757, opened 2022-04-27
Problem: ibc solo machine integration not tested (fix #613)

### LINKED ISSUE #613: Problem: ibc solo machine integration not tested
perhaps after some 0.1 release is tagged: https://github.com/crypto-com/ibc-solo-machine ?
cc @MaCong-crypto

linfeng-crypto:


integration_tests/test_solo_machine.py
```diff
@@ -0,0 +1,182 @@
...
+        file = tarfile.open(tar_file)
+        file.extractall(self.bin_dir)
+        file.close()
+        # check file exists
+        if not os.path.exists(self.bin_file):
+            raise Exception(f"solomachine binary file {self.bin_file} is not exists")
+        if not os.path.exists(self.sign_file):
+            raise Exception(
+                f"solomachine signature file {self.sign_file} is not exists"
+            )
```
tomtau, line 105: probably ok, but it seems a bit cumbersome -- it can be replaced by having it fetched in the Nix environment (`builtins.fetchTarball`) and having the solo machine binary in the environment's PATH

integration_tests/test_solo_machine.py
```diff
@@ -0,0 +1,182 @@
...
+            url = "https://github.com/crypto-com/ibc-solo-machine/releases/\
+download/v0.1.1/macos-latest-v0.1.1.tar.gz"
+            self.sign_file = os.path.join(self.bin_dir, "libmnemonic_signer.dylib")
+        else:
+            url = "https://github.com/crypto-com/ibc-solo-machine/releases/\
+download/v0.1.1/ubuntu-latest-v0.1.1.tar.gz"
+            self.sign_file = os.path.join(self.bin_dir, "libmnemonic_signer.so")
+        local_filename = url.split("/")[-1]
+        tar_file = f"/tmp/{local_filename}"
+        if os.path.exists(self.bin_file) and os.path.exists(self.sign_file):
```
tomtau, line 88: this should probably check the sha256 checksum of the binaries as well

integration_tests/test_solo_machine.py
```diff
@@ -0,0 +1,182 @@
...
+    balance_1 = int(solo_machine.get_balance())
+    assert balance_1 == 0x20
+    time.sleep(3)
+    data = solo_machine.burn("0x10")
+    assert data["result"] == "success"
+    assert data["data"]["amount"] == "0x10"
+    time.sleep(3)
+    balance_2 = int(solo_machine.get_balance())
+    assert balance_2 == balance_1 - 0x10
+    assert balance_1 > balance_2
```
tomtau, line 182: some of these assertions should be against the node directly, i.e. to check that the chain/cluster node indeed received it (if there's a bug in the solo machine CLI or if it connects to a wrong node)

tomtau (changes requested):
looks all right, just two fixes to consider: 1) making the binary download more robust (e.g. by using Nix), 2) additional assertions against the node in the cluster

integration_tests/test_solo_machine.py
tomtau: (or use Nix which will do it by default :))

devashishdxt (changes requested):
Can you change the PR to follow the [C4 Spec](https://rfc.zeromq.org/spec/42/)?

> A patch commit message MUST consist of a single short (less than 50 characters) line stating the problem (“Problem: …") being solved, followed by a blank line and then the proposed solution (“Solution: …").

integration_tests/test_solo_machine.py
linfeng-crypto: sure

AlphaMale1st (approved):

integration_tests/test_solo_machine.py
```diff
@@ -0,0 +1,179 @@
...
+
+    # fixme after this PR merged, currently only support hex string amount
+    # https://github.com/crypto-com/ibc-solo-machine/pull/82
+    def mint(self, amount="0x20"):
+        sub_cmd = f"ibc mint {self.chain_id} {amount} solotoken"
+        return self.run_sub_cmd(sub_cmd)
+
+    # fixme fix after this PR merged, currently only support hex string amount
+    # https://github.com/crypto-com/ibc-solo-machine/pull/82
+    def burn(self, amount="0x10"):
```
tomtau, line 125: there's now https://github.com/crypto-com/ibc-solo-machine/releases/tag/v0.1.2 , so it can be fixed?

linfeng-crypto: sure

.github/workflows/nix.yml
```diff
@@ -99,6 +99,28 @@ jobs:
...
+        run: make nix-integration-test-solomachine
+      - name: Tar debug files
+        if: failure()
+        run: tar cfz debug_files_solomachine.tar.gz -C /tmp/pytest-of-runner .
+      - uses: actions/upload-artifact@v2
+        if: failure()
+        with:
+          name: debug_files_solomachine
+          path: debug_files_solomachine.tar.gz
+          if-no-files-found: ignore
```
tomtau, line 122: maybe ok, but are these debug files necessary?

integration_tests/test_solo_machine.py
```diff
@@ -0,0 +1,179 @@
...
+        os.environ["SOLO_HD_PATH"] = "m/44'/394'/0'/0/0"
+        os.environ["SOLO_ACCOUNT_PREFIX"] = "cro"
+        os.environ["SOLO_ADDRESS_ALGO"] = "secp256k1"
+        os.environ["SOLO_FEE_AMOUNT"] = "1000"
+        os.environ["SOLO_FEE_DENOM"] = "basecro"
+        os.environ["SOLO_GAS_LIMIT"] = "300000"
+        os.environ["SOLO_GRPC_ADDRESS"] = f"http://0.0.0.0:{self.grpc_port}"
+        os.environ["SOLO_RPC_ADDRESS"] = f"http://0.0.0.0:{self.rpc_port}"
+        os.environ["SOLO_TRUSTED_HASH"] = self.trusted_hash
+        os.environ["SOLO_TRUSTED_HEIGHT"] = self.trusted_height
```
tomtau, line 84: could this be taken from pystarport's yaml config instead of hardcoded here?

integration_tests/test_solo_machine.py
```diff
@@ -0,0 +1,179 @@
...
+        return self.run_sub_cmd(sub_cmd)["data"]["balance"]
+
+    # fixme after this PR merged, currently only support hex string amount
+    # https://github.com/crypto-com/ibc-solo-machine/pull/82
+    def mint(self, amount="0x20"):
+        sub_cmd = f"ibc mint {self.chain_id} {amount} solotoken"
+        return self.run_sub_cmd(sub_cmd)
+
+    # fixme fix after this PR merged, currently only support hex string amount
+    # https://github.com/crypto-com/ibc-solo-machine/pull/82
```
tomtau, line 124: the solo machine was upgraded to 0.1.2, so these comments can be removed and amounts can be passed in a decimal format?

integration_tests/test_solo_machine.py
```diff
@@ -0,0 +1,179 @@
...
+    """
+    check solo machine
+    """
+    cli = cluster
+    wait_for_block(cli, 1)
+
+    # get the chain balance
+    chain_solo_addr = cluster.address("solo-signer")
+    chain_balance_1 = cluster.balance(chain_solo_addr)
+    assert chain_balance_1 == 1500 * 10 ** 8
```
tomtau, line 140: the magic number can be extracted into a constant?  e.g. `CRO_DECIMALS = 10 ** 8` unless it's already defined somewhere else?

integration_tests/test_solo_machine.py
```diff
@@ -0,0 +1,179 @@
...
+        return self.run_sub_cmd(sub_cmd)
+
+    def connect_chain(self):
+        sub_cmd = f"ibc connect {self.chain_id}"
+        result = self.run_sub_cmd(sub_cmd, False)
+        if "error" in result:
+            raise Exception(result)
+
+    def get_balance(self):
+        sub_cmd = f"chain balance {self.chain_id} solotoken"
```
tomtau, line 114: I guess "solotoken" is the denomination? Can it be extracted into a constant or an argument/config parameter?

integration_tests/test_solo_machine.py
```diff
@@ -0,0 +1,179 @@
...
+    # mint
+    data = solo_machine.mint(20)
+    assert data["result"] == "success"
+    assert int(data["data"]["amount"], 16) == 20
+    wait_for_block(cli, 2)
+    balance_1 = int(solo_machine.get_balance())
+    assert balance_1 == 20
+
+    # check the chain balance
+    chain_balance_2 = cli.balance(chain_solo_addr)
```
tomtau, line 166: besides checking the on-chain balance, it'll be good to assert the denomination is correct as well as well

tomtau (commented):
the binary setup looks all right now.
there are two minor things to sort out:
1) not hardcoding the environment config in the test
2) checking and asserting on-chain denominations

I assume you've read [contribution guidelines](https://github.com/crypto-org-chain/chain-main/blob/master/CONTRIBUTING.md#working-on-issues), so ideally you can also squash your commits after the changes and [adjust the commit message](https://github.com/crypto-org-chain/chain-main/pull/757#pullrequestreview-954786997) as requested before.

AlphaMale1st (approved):

.github/workflows/nix.yml
linfeng-crypto: I think so, same as other test settings.

integration_tests/test_solo_machine.py
```diff
@@ -0,0 +1,182 @@
...
+        self.trusted_height = result["result"]["block"]["header"]["height"]
+
+    def set_env(self):
+        os.environ["SOLO_DB_URI"] = self.db_path
+        os.environ["SOLO_SIGNER"] = self.sign_file
+        os.environ[
+            "SOLO_MNEMONIC"
+        ] = "awesome there minute cash write immune tag \
+reopen price congress trouble reunion south wisdom donate credit below leave \
+wisdom eagle sail siege rice train"
```
tomtau, line 81: why duplicate it in the test code and not take it from the test config ("solo_machine.yaml")?

tomtau (changes requested):
could you rebase against the latest master, given it's updated the ibc-go module?

integration_tests/test_solo_machine.py
linfeng-crypto: I'll fix it.

tomtau (approved):
will merge, the other nits can be addressed in a future PR

PR #763, opened 2022-05-12
Problem: Relayer incentivization is not enabled in `chain-main`

devashishdxt:
Added relayer incentivization code in `app.go`. Related to #765.


app/app.go
```diff
@@ -357,14 +365,26 @@ func New(
 		),
 	)
 
+	// IBC Fee Module keeper
+	app.IBCFeeKeeper = ibcfeekeeper.NewKeeper(
+		appCodec, keys[ibcfeetypes.StoreKey], app.GetSubspace(ibcfeetypes.ModuleName),
+		app.IBCFeeKeeper, // may be replaced with IBC middleware
```
tomtau, line 371: so IBC middleware isn't there yet?

go.mod
```diff
@@ -4,8 +4,8 @@ go 1.17
 
 require (
 	github.com/confluentinc/bincover v0.1.0
-	github.com/cosmos/cosmos-sdk v0.45.3
-	github.com/cosmos/ibc-go/v3 v3.0.0
+	github.com/cosmos/cosmos-sdk v0.45.4
+	github.com/cosmos/ibc-go/v3 v3.0.0-alpha1.0.20220510152759-dcd0681d8f07
```
tomtau, line 8: I guess one can't put `ics29-beta2` as the version string?

x/icaauth/module_ibc.go
```diff
@@ -38,13 +40,17 @@ func (am IBCModule) OnChanOpenInit(
 	chanCap *capabilitytypes.Capability,
 	counterparty channeltypes.Counterparty,
 	version string,
-) error {
+) (string, error) {
```
tomtau, line 43: is icaauth ok with future channel upgrades?

tomtau (approved):
looks all right; one thing to consider is if we know how that channel upgrade functionality will work and whether there's anything to change in ICA auth to prepare it for it

app/app.go
devashishdxt: We can add more middlewares in the future. I'll modify the comment.

go.mod
devashishdxt: Correct. It's just a version tag and not a release, so, `go mod tidy` automatically replaces it with this version.

x/icaauth/module_ibc.go
devashishdxt: I don't think there are any changes required in `chain-main`. But, I think, we'll need some changes in `ibc-go` ([here](https://github.com/cosmos/ibc-go/blob/9f70a070d773f8bfdb62d4205d8878f3149f351a/modules/apps/27-interchain-accounts/controller/keeper/account.go#L45)). I've asked Carlos about this.

PR #767, opened 2022-05-16
Problem: ICA test fails on `master` branch

### LINKED ISSUE #766: Problem: ica test fails on main
https://github.com/crypto-org-chain/chain-main/runs/6445725049?check_suite_focus=true

devashishdxt:
Solution: Temporarily disable fee middleware for ICA keepers. Fixes #766.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -405,14 +405,20 @@ func New(
 
 	icaAuthModule := icaauthmodule.NewAppModule(appCodec, app.ICAAuthKeeper)
 
+	// TODO: Remove this `nolint` when fee middleware is added
+	// nolint: gosimple
 	var icaAuthStack porttypes.IBCModule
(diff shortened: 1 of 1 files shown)


tomtau (approved):
so the root cause was the ICS29 middleware over ICA? is there a bug in ibc-go for it?

devashishdxt:
> so the root cause was the ICS29 middleware over ICA? is there a bug in ibc-go for it?

This doesn't happen with latest `hermes`. I'll try and upgrade `hermes` version used in integration tests and add it back in future PRs.

devashishdxt:
Anyway, with current code for ICA controller keeper, we cannot create fee enabled channels for ICA accounts. I've asked Carlos about this. Waiting for his reply before creating an issue in `ibc-go`.

PR #771, opened 2022-05-19
Problem: Cannot build testnet binaries in CI

devashishdxt:
Solution: Changed the build flags in testnet branch

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -15,7 +15,7 @@ builds:
     goarch:
       - amd64
     flags:
-        - -tags=cgo,ledger,!test_ledger_mock,!ledger_mock,!ledger_zemu
+        - -tags=cgo,ledger,!test_ledger_mock,!ledger_mock,!ledger_zemu,testnet
     ldflags:
       - -s -w -X github.com/cosmos/cosmos-sdk/version.Name=crypto-org-chain -X github.com/cosmos/cosmos-sdk/version.AppName=chain-maind -X github.com/cosmos/cosmos-sdk/version.Version={{.Version}} -X github.com/cosmos/cosmos-sdk/version.Commit={{.Commit}}
   - id: "chain-maind-darwin-arm64"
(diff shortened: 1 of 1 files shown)


.goreleaser.yml
```diff
@@ -15,7 +15,7 @@ builds:
     goarch:
       - amd64
     flags:
-        - -tags=cgo,ledger,!test_ledger_mock,!ledger_mock,!ledger_zemu
+        - -tags=cgo,ledger,!test_ledger_mock,!ledger_mock,!ledger_zemu,testnet
```
yihuang, line 18: aren't we need to create a new bunch of builds for the testnet, rather than modify the mainnet ones? otherwise, you'll need to change it back?

devashishdxt: This commit is only made on `crypto-org-chain:release/v4-croeseid` and not on `master` branch.

devashishdxt: Other option is to create both mainnet and testnet binaries for all tags?

yihuang: > Other option is to create both mainnet and testnet binaries for all tags?

that's what cronos repo do, so we don't need to create different tags for networks.

devashishdxt: @tomtau Should we do the same for `chain-main` as well?

yihuang: maybe we can discuss this in another place, since we already have the testnet release branch, we can do this for now I think.

yihuang (approved):

.goreleaser.yml
tomtau: here historically, the testnet has been long-running and the network existence diverged from the mainnet... while the mainnet network upgrade path is v1 (0.42)->v2 (0.42+NFT)->v3 (0.44), the long-running testnet has been on pre-v3 (0.33) -> v3 (0.44).

For Cronos, the reason of not doing the separate releases was that they were not worth the effort: 1) the release cadence was higher, 2) the public testnet lifetime was often short (sometimes after 1-2 weeks, a new network from scratch had to be done), 3) the public testnet (more or less) matches the mainnet and doesn't need a custom upgrade path.

Anyway, I told @devashishdxt it's not set in stone here and perhaps testnet build tags could be eliminated altogether and add the runtime (`--testnet` option); if there's something to be changed, @devashishdxt perhaps can draft ADR for the repo.

For the upcoming testnet/these currently released binaries, here are a few things to consider:

1. it's "alpha"/"preview" -- i.e. their exact code / dependencies won't be on mainnet + the public testnet may again have its own custom upgrade path to a stable release (that may be similar to future mainnet)
2. the existing public testnet will likely be decommissioned and a new one will be started with these binaries (so no need to worry about that pre-v3 -> v3 -> pre-v4)

PR #780, opened 2022-05-30
Problem: CosmWasm is not enabled on `chain-main`

devashishdxt:
Solution: Integrated `wasmd` in `app.go` (have used temporary forks for ics29 fee middleware changes)


go.mod
```diff
@@ -143,4 +146,10 @@ replace github.com/opencontainers/image-spec => github.com/opencontainers/image-
...
 replace github.com/opencontainers/runc => github.com/opencontainers/runc v1.0.3
 
-replace github.com/tecbot/gorocksdb => github.com/cosmos/gorocksdb v1.2.0
+// replace github.com/tecbot/gorocksdb => github.com/cosmos/gorocksdb v1.2.0
+
+replace github.com/cosmos/ibc-go/v3 => github.com/cosmos/ibc-go/v3 v3.0.0-alpha1.0.20220510152759-dcd0681d8f07
+
+replace github.com/CosmWasm/wasmd => github.com/devashishdxt/wasmd v0.27.1-0.20220530042031-4126e5b964df
+
+replace github.com/CosmWasm/wasmvm => github.com/yihuang/wasmvm v1.0.1-0.20220601153132-00fa4d9fdb62
```
yihuang, line 155: https://github.com/CosmWasm/wasmvm/pull/335

needed for building binaries that don't rely on build-time environments.

yihuang:
The [windows build doesn't work](https://github.com/crypto-org-chain/chain-main/runs/6694898385?check_suite_focus=true) because no prebuilt libwasmvm.dll in wasmvm repo.

chrisntfs0711 (approved):

tomtau:
Would it be possible to fix the Windows build? I think one can just add a `cargo build` step to this batch file: https://github.com/crypto-org-chain/chain-main/blob/master/makewin.bat to build wasmvm and have it on the path?

default.nix
```diff
@@ -60,7 +62,11 @@ let
       mv ./${instrumentedBinary} $instrumented/bin/
     '';
     preFixup = ''
-      find $instrumented/bin/ -type f 2>/dev/null | xargs -r remove-references-to -t ${pkgs.go} || true
+      find $instrumented/bin/ -type f 2>/dev/null | xargs -r remove-references-to -t ${go} || true
+    '';
+    postFixup = pkgs.lib.optionalString pkgs.stdenv.isDarwin ''
+      install_name_tool -change "@rpath/libwasmvm.dylib" ${pkgs.libwasmvm}/lib/libwasmvm.dylib $out/bin/chain-maind
+      install_name_tool -change "@rpath/libwasmvm.dylib" ${pkgs.libwasmvm}/lib/libwasmvm.dylib $instrumented/bin/chain-maind-inst
```
yihuang, line 69:
```suggestion
```

we can revert these lines now.

app/app.go
```diff
@@ -478,15 +559,15 @@ func New(
 		slashingtypes.ModuleName, evidencetypes.ModuleName, stakingtypes.ModuleName, ibchost.ModuleName,
 		ibctransfertypes.ModuleName, authtypes.ModuleName, banktypes.ModuleName, govtypes.ModuleName,
 		genutiltypes.ModuleName, authz.ModuleName, feegrant.ModuleName, paramstypes.ModuleName, vestingtypes.ModuleName,
-		icatypes.ModuleName, icaauthmoduletypes.ModuleName, ibcfeetypes.ModuleName, chainmaintypes.ModuleName,
-		nfttypes.ModuleName, supplytypes.ModuleName,
+		icatypes.ModuleName, icaauthmoduletypes.ModuleName, wasm.ModuleName, ibcfeetypes.ModuleName,
```
tomtau, line 562: I saw "wasm" was the last in wasmd's app.go, not sure whether required or ok

app/app.go
```diff
@@ -559,6 +643,18 @@ func New(
...
+		err = manager.RegisterExtensions(
+			wasmkeeper.NewWasmSnapshotter(app.CommitMultiStore(), &app.WasmKeeper),
+		)
+		if err != nil {
+			panic(fmt.Errorf("failed to register snapshot extension: %s", err))
+		}
+	}
+
 	planName := "v4.0.0"
 	app.UpgradeKeeper.SetUpgradeHandler(planName, func(ctx sdk.Context, _ upgradetypes.Plan, fromVM module.VersionMap) (module.VersionMap, error) {
```
tomtau, line 659: probably a TODO to set wasm-related items in the upgrade handler?

go.sum
```diff
@@ -223,6 +229,7 @@ github.com/coreos/pkg v0.0.0-20160727233714-3ac0863d7acf/go.mod h1:E3G3o1h8I7cfc
 github.com/coreos/pkg v0.0.0-20180928190104-399ea9e2e55f/go.mod h1:E3G3o1h8I7cfcXa63jLwjI0eiQQMgzzUDFVpN/nH/eA=
 github.com/cosmos/btcutil v1.0.4 h1:n7C2ngKXo7UC9gNyMNLbzqz7Asuf+7Qv4gnX/rOdQ44=
 github.com/cosmos/btcutil v1.0.4/go.mod h1:Ffqc8Hn6TJUdDgHBwIZLtrLQC1KdJ9jGJl/TvgUaxbU=
+github.com/cosmos/cosmos-sdk v0.45.1/go.mod h1:XXS/asyCqWNWkx2rW6pSuen+EVcpAFxq6khrhnZgHaQ=
```
tomtau, line 232: is it from wasmd?

tomtau (commented):
I was trying to compare with the wasmd's app.go wiring... there are a few missing parts, probably not critical:
- ICA https://github.com/CosmWasm/wasmd/blob/v0.27.0/app/app.go#L485-L492 -- it mentions it's experimental, so perhaps better to omit?
- https://github.com/CosmWasm/wasmd/blob/v0.27.0/app/app.go#L765-L768 -- `InitializePinnedCodes` -- not sure if needed?
- there may be others, but I guess the most important ones were the two extra Ante handlers and there were added

Also, I guess the binary release via GoReleaser may not work any more?

devashishdxt:
> * ICA https://github.com/CosmWasm/wasmd/blob/v0.27.0/app/app.go#L485-L492 -- it mentions it's experimental, so perhaps better to omit?

This is their sample `icaauth` module. We already have wiring for that in `app.go`.

app/app.go
devashishdxt: They mention that it only needs to come after IBC transfer module. In our case, I've put it after all the cosmos and ibc modules.

go.sum
devashishdxt: No. This is in `go.sum` because we have `ibc-go v3.0.0` in `go.mod` but we replace it with `ics29-beta2`. So this is not getting used. `go mod graph` also doesn't show this getting used.

devashishdxt:
> * https://github.com/CosmWasm/wasmd/blob/v0.27.0/app/app.go#L765-L768 -- `InitializePinnedCodes` -- not sure if needed?

Added

devashishdxt:
> Also, I guess the binary release via GoReleaser may not work any more? (at least, I assume it won't bundle the wasmvm dynamic library built by Rust by default)

As far as I understand this, pre-built dynamic libraries for linux and mac are linked by [`wasmvm`](https://github.com/CosmWasm/wasmvm). For windows, we may have to include `.dll` until https://github.com/CosmWasm/wasmvm/pull/288 is merged and released. I've asked a question on discord for more details on this.

devashishdxt:
> Would it be possible to fix the Windows build? I think one can just add a `cargo build` step to this batch file: https://github.com/crypto-org-chain/chain-main/blob/master/makewin.bat to build wasmvm and have it on the path?

I'll do this in next PR.

go.mod
```diff
@@ -143,4 +146,10 @@ replace github.com/opencontainers/image-spec => github.com/opencontainers/image-
 
 replace github.com/opencontainers/runc => github.com/opencontainers/runc v1.0.3
 
-replace github.com/tecbot/gorocksdb => github.com/cosmos/gorocksdb v1.2.0
+// replace github.com/tecbot/gorocksdb => github.com/cosmos/gorocksdb v1.2.0
+
+replace github.com/cosmos/ibc-go/v3 => github.com/cosmos/ibc-go/v3 v3.0.0-alpha1.0.20220510152759-dcd0681d8f07
+
+replace github.com/CosmWasm/wasmd => github.com/devashishdxt/wasmd v0.27.1-0.20220530042031-4126e5b964df
```
tomtau, line 153: could you update it to this fork: https://github.com/crypto-org-chain/wasmd ?

tomtau (changes requested):
updating the fork -- the windows build is currently marked as "Required" workflow to pass, so not sure if this can be merged without it being fixed

.gitmodules
```diff
@@ -1,3 +1,7 @@
 [submodule "cosmos-sdk"]
 	path = third_party/cosmos-sdk
 	url = https://github.com/cosmos/cosmos-sdk.git
+
+[submodule "wasmd"]
```
yihuang, line 5: Why do we need the submodule?

devashishdxt: Simulation tests depend on some `wasm` files from repository.

go.mod
devashishdxt: Done

.gitmodules
devashishdxt: https://github.com/crypto-org-chain/chain-main/pull/780/files#diff-6377d84e477b544abf9150c43e9f581ac6aaa209cee8f26961187cebe0d200a0

yihuang: the `reflect.wasm` contract, if that's the only thing we need, maybe better to just copy it?

devashishdxt: Then we'll have to manually manage the file with every release of `wasmd`.

yihuang:
> ❌ `/runsim` simulation test has failed 😅 Can further check [here](https://github.com/crypto-org-chain/chain-main/actions/runs/2486568576)

it failed, should we run again?

devashishdxt:
> > ❌ `/runsim` simulation test has failed 😅 Can further check [here](https://github.com/crypto-org-chain/chain-main/actions/runs/2486568576)
> 
> 
> 
> it failed, should we run again?

Yes. We can run again. It failed because it didn't find `runsim` binary.

yihuang (approved):
LGTM

.github/workflows/buildwin.yml
```diff
@@ -15,13 +16,21 @@ jobs:
...
+          toolchain: stable
       - name: Normal check out code
         uses: actions/checkout@v2
         with:
           submodules: true
       - name: Set GOBIN
         run: |
           echo "$(go env GOPATH)/bin" >> $GITHUB_PATH
+      - name: Build wasmvm
+        working-directory: ./third_party/wasmvm/libwasmvm
```
yihuang, line 32: I think we can download a released source tarball of libwasmvm on the fly instead of maintaining a submodule.

devashishdxt: They don't have released version for windows. Anyway, I don't think we need to build `libwasm` to test windows build. We can just build `chainmain` with `sys_wasmvm` flag to test build. I'm trying it out. What do you think?

devashishdxt: For release, we'll have to package it together.

tomtau: that would be possible after https://github.com/CosmWasm/wasmvm/pull/288

tomtau: or do you mean to publish a tarball in our own fork in the meantime?

yihuang: I mean, sth like this:
```
curl -OL https://github.com/CosmWasm/wasmvm/archive/refs/tags/v1.0.0.tar.gz
tar xfz ...
cargo build ...
```

devashishdxt: We've forked `wasmvm`. No need for submodule anymore.

Issue #781, opened 2022-06-02
Problem: released binary don't contain needed dynamic libraries


yihuang:
After [integrating cosmwasm](https://github.com/crypto-org-chain/chain-main/pull/780), the `chain-maind` binary will require the `libwsamvm.[so|dylib]` to run, it would be good to include the dynamic libraries inside the released bundle.

Solution:
- Adopt the [nix bundle solution used in cronos](https://github.com/crypto-org-chain/cronos/pull/476)

- [x] #793 linux, x86-64 mac
- [ ] TODO arm mac
- [ ] TODO windows

tomtau:
will also need to bundle the Windows DLL

devashishdxt:
We don't need this anymore. Closing.

PR #793, opened 2022-06-28
Problem: released binary don't contain needed dynamic libraries

yihuang:
WIP: #781
Solution:
- Support Linux and x86-64 mac.
- flakify chain-maind project
- port nix releaser from cronos project

👮🏻👮🏻👮🏻 !!!! REFERENCE THE PROBLEM YOUR ARE SOLVING IN THE PR TITLE AND DESCRIBE YOUR SOLUTION HERE !!!! DO NOT FORGET !!!! 👮🏻👮🏻👮🏻

# PR Checklist:

- [ ] Have you read the [CONTRIBUTING.md](https://github.com/crypto-org-chain/chain-main/blob/master/CONTRIBUTING.md)?
- [ ] Have you rebased your work on top of the latest master? 
- [ ] Have you checked your code compiles? (`make`)
- [ ] Have you included tests for any non-trivial functionality?
- [ ] If you are contributing for the first time, please read the agreement in [CONTRIBUTING.md](https://github.com/crypto-org-chain/chain-main/blob/master/CONTRIBUTING.md) now and add a comment to this pull request stating that your PR is in accordance with the [Developer's Certificate of Origin](https://github.com/crypto-org-chain/chain-main/blob/master/CONTRIBUTING.md#developer-certificate-of-originn).


tomtau (commented):
maybe the RocksDB builds can be removed? they are not essential here, so the release script could be simplified?

yihuang:
> maybe the RocksDB builds can be removed? they are not essential here, so the release script could be simplified?

and should remove the testnet binaries too I guess?

tomtau:
> > maybe the RocksDB builds can be removed? they are not essential here, so the release script could be simplified?
> 
> and should remove the testnet binaries too I guess?

Perhaps, that's still pending a discussion: @devashishdxt suggested before it may be preferable to release both at the same time: https://github.com/crypto-org-chain/chain-main/pull/771#discussion_r876564455

yihuang:
> > > maybe the RocksDB builds can be removed? they are not essential here, so the release script could be simplified?
> > 
> > 
> > and should remove the testnet binaries too I guess?
> 
> Perhaps, that's still pending a discussion: @devashishdxt suggested before it may be preferable to release both at the same time: [#771 (comment)](https://github.com/crypto-org-chain/chain-main/pull/771#discussion_r876564455)

I'll remove from the release workflow for now, but keep in the nix build matrix, it'll be trivial to add later.

yihuang:
release workflow test run: https://github.com/yihuang/chain-main/releases/tag/v0.8.0-test

yihuang:
https://github.com/yihuang/chain-main/runs/7104981504?check_suite_focus=true

the arm Linux version took 2h+ to build, it's built with an emulator.

tomtau:
does it need an emulator? normally arm64 toolchain has x64 binaries for cross-compiling

yihuang:
> does it need an emulator? normally arm64 toolchain has x64 binaries for cross-compiling

currently, I find the emulator(binfmt wrapper) solution is the most effortless one, works out of the box. cross-compiling solution has some issues with go compiler the last time I try, but takes some time to investigate.

.github/workflows/release.yml
```diff
@@ -0,0 +1,143 @@
...
+          echo 'EOF' >> $GITHUB_ENV
+          cat $GITHUB_ENV
+      - name: upload linux arm64 binaries
+        uses: softprops/action-gh-release@v1
+        with:
+          draft: true
+          files: "${{ env.FILES }}"
+
+  nix-bundle-x86_64-windows:
+    if: ${{ false }}
```
yihuang, line 84: skipped for now

.github/workflows/release.yml
```diff
@@ -0,0 +1,147 @@
...
+          ls -1 checksums-*.txt >> $GITHUB_ENV
+          echo 'EOF' >> $GITHUB_ENV
+          cat $GITHUB_ENV
+      - name: upload linux/mac x86_64 binaries
+        uses: softprops/action-gh-release@v1
+        with:
+          draft: true
+          files: "${{ env.FILES }}"
+
+  nix-bundle-aarch64-linux:
```
tomtau, line 49: what about aarch64-macos?

.github/workflows/release.yml
```diff
@@ -0,0 +1,147 @@
...
+        uses: actions/checkout@v2
+        with:
+          submodules: true
+      - name: setup msys2
+        uses: msys2/setup-msys2@v2
+        with:
+          msystem: MINGW64          
+          update: false
+          install: git msys2-devel base-devel mingw-w64-x86_64-gcc mingw-w64-x86_64-rocksdb  mingw-w64-x86_64-go  mingw-w64-x86_64-jq  mingw-w64-x86_64-dlfcn        
+      - name: get rocksdb & build
```
tomtau, line 103: no need to build rocksdb

.github/workflows/release.yml
```diff
@@ -0,0 +1,147 @@
...
+          cp /mingw64/bin/libsnappy.dll ./build
+          cp /mingw64/bin/libstdc++-6.dll ./build
+          cp /mingw64/bin/libwinpthread-1.dll ./build
+          cp /mingw64/bin/libzstd.dll ./build
+          cp /mingw64/bin/zlib1.dll ./build
+          cd ./build
+          tar cvfz ../chain-main.tar.gz *
+          cd ..
+          ls -la ./chain-main.tar.gz
+          mv ./chain-main.tar.gz chain-main_${GITHUB_REF_NAME:1}-rocksdb_${PLATFORM}.tar.gz
```
tomtau, line 134: it doesn't seem to build or fetch the CosmWasm DLL?

default.nix
```diff
@@ -20,188 +32,62 @@ let
...
-          --format="::error file=%(path)s,line=%(row)d,col=%(col)d::%(path)s:%(row)d:%(col)d: %(code)s %(text)s" \
-          || EXIT_STATUS=$?
-        find . -name "*.nix" -type f | xargs nixpkgs-fmt --check || EXIT_STATUS=$?
-        exit $EXIT_STATUS
-      '')
-    ];
+buildGoApplication rec {
+  pname = "chain-maind";
+  version = "4.0.0";
+  go = go_1_17;
```
tomtau, line 39: should it be 1.17 and not 1.18?

tomtau (commented):
those integration test changes are due to changes in linting rules after the tooling upgrade?
I guess QEMU is the option if the cross-compiling is troublesome and there isn't a runner support yet: https://github.community/t/ask-about-github-hosted-runners-arm64-platform-support/167483/5

Two things that seem to be missing: 1) Windows release builds (the current flow doesn't seem correct), 2) arm64 macOS (or does that bundle flow create the "universal" binary?)

.github/workflows/release.yml
tomtau: oh it's skipped: https://github.com/crypto-org-chain/chain-main/pull/793#discussion_r909273757

yihuang:
> those integration test changes are due to changes in linting rules after the tooling upgrade?

yes, the upgraded `nixpkgs` come with a newer version of python, which breaks some python dependency libraries.

.github/workflows/release.yml
yihuang: I removed the windows job for now.

.github/workflows/release.yml
yihuang: I removed the windows job for now.

.github/workflows/release.yml
yihuang: I updated the issue description to separate it into three steps, let's support arm mac and windows in another PR.

.github/workflows/release.yml
yihuang: I updated the issue description to separate it into three steps, let's support arm mac and windows in another PR.

.github/workflows/release.yml
yihuang: I updated the issue description to separate it into three steps, let's support arm mac and windows in another PR.

default.nix
yihuang: done

tomtau (approved):

PR #805, opened 2022-07-08
Problem: Missing upgrade handler for wasmd

### LINKED ISSUE #785: Problem: missing upgrade handlers for wasmd
https://github.com/crypto-org-chain/chain-main/blob/master/doc/architecture/adr-001.md#decision
> the upgrade handler should be added to initialize the module with the setting similar to the "gaia-flex" network parameters:

devashishdxt:
Solution: Added an upgrade handler. Fixes #785.


yihuang (approved):
LGTM

PR #806, opened 2022-07-11
Problem: Changelog is not updated

devashishdxt:
Solution: Updated changelog


tomtau (approved):
good to add dates for those tagged (pre-)releases?

devashishdxt:
> good to add dates for those tagged (pre-)releases?

Good idea. Lets start adding dates with all the future releases.

PR #815, opened 2022-07-15
Problem: Jailed validators can sometimes become unbonded before assert statement is executed in integration tests

devashishdxt:
Solution: Assert either unbonding or unbonded status


tomtau (approved):

yihuang (approved):
LGTM

PR #817, opened 2022-07-18
Problem: Release script does not create testnet binaries on `release/v4-croeseid` branch

devashishdxt:
Solution: Changed the release script to create testnet binaries

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -68,7 +68,7 @@ jobs:
           BUILD_TYPE="tarball"
           FLAKE="github:${GITHUB_REPOSITORY}/${GITHUB_REF_NAME}"
 
-          nix build -L --extra-platforms aarch64-linux ${FLAKE}#packages.aarch64-linux.chain-maind-$BUILD_TYPE
+          nix build -L --extra-platforms aarch64-linux ${FLAKE}#packages.aarch64-linux.chain-maind-testnet-$BUILD_TYPE
           cp result chain-main_${GITHUB_REF_NAME:1}_${PLATFORM}.tar.gz
 
           sha256sum *.tar.gz > checksums-$PLATFORM.txt
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,7 +29,7 @@ jobs:
           BUILD_TYPE="tarball"
           FLAKE="github:${GITHUB_REPOSITORY}/${GITHUB_REF_NAME}"
 
-          nix build -L ${FLAKE}#chain-maind-$BUILD_TYPE
+          nix build -L ${FLAKE}#chain-maind-testnet-$BUILD_TYPE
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -29,7 +29,7 @@ jobs:
           BUILD_TYPE="tarball"
           FLAKE="github:${GITHUB_REPOSITORY}/${GITHUB_REF_NAME}"
 
-          nix build -L ${FLAKE}#chain-maind-$BUILD_TYPE
+          nix build -L ${FLAKE}#chain-maind-testnet-$BUILD_TYPE
           cp result chain-main_${GITHUB_REF_NAME:1}_${PLATFORM}.tar.gz
```
tomtau, line 33: is this correct?

tomtau (commented):
will this also build only the goleveldb binaries?

tomtau:
arm64 linux may need a patch as well: https://github.com/crypto-org-chain/chain-main/blob/8c05c8bc1f5430618ba4ebe13fa185998c265dad/.github/workflows/release.yml#L71 ?

.github/workflows/release.yml
devashishdxt: Yes. The DB backend defaults to `goleveldb`.

devashishdxt:
> arm64 linux may need a patch as well:
> 
> https://github.com/crypto-org-chain/chain-main/blob/8c05c8bc1f5430618ba4ebe13fa185998c265dad/.github/workflows/release.yml#L71
> 
> ?

Changed

tomtau (approved):

devashishdxt:
I'll wait for @yihuang to review this before merging.

.github/workflows/release.yml
yihuang: The asset name should has the surfix too: `chain-main_4.0.0-alpha3-croeseid_Darwin_arm64_croeseid.tar.gz`

devashishdxt: @allthatjazzleo Can you check the final asset name? I remember you removed the `_croeseid` postfix in https://github.com/crypto-org-chain/chain-main/pull/776 and all the automation scripts assume that there's no `_croeseid` postfix.

allthatjazzleo: @yihuang @devashishdxt 
according to our previous convention, we don't have `testnet` suffix
https://github.com/crypto-org-chain/chain-main/releases/tag/v3.1.1-croeseid

devashishdxt: @allthatjazzleo Final generated assets will look like this: https://github.com/devashishdxt/chain-main/releases/tag/untagged-cab15c548893a1949350

allthatjazzleo: > @allthatjazzleo Can you check the final asset name? I remember you removed the `_croeseid` postfix in #776 and all the automation scripts assume that there's no `_croeseid` postfix.

yes coz croeseid is already part of the version tag

yihuang (approved):
LGTM

PR #818, opened 2022-07-22
Problem: License file is not up-to-dated

calvinaco:
Solution: Update LICENSE file

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-  Copyright 2018 - present Crypto.org
+  Copyright 2018-present Crypto.org
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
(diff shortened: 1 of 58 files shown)


tomtau (changes requested):
the copyright is missing the copyright holder information, as per the Apache2 template: https://www.apache.org/licenses/LICENSE-2.0
```
Copyright [yyyy] [name of copyright owner]
```

calvinaco:
> the copyrigh statementt is missing the copyright holder information, as per the Apache2 template (see "HOW TO APPLY THE APACHE LICENSE TO YOUR WORK"): https://www.apache.org/licenses/LICENSE-2.0
> 
> ```
> Copyright [yyyy] [name of copyright owner]
> ```

Updated, thanks!

tomtau (approved):
there are also notices in the modified files in x/nft -- I guess they should be updated as well?

calvinaco:
> there are also notices in the modified files in x/nft -- I guess they should be updated as well?

Updated, there are some differences in the year between the LICENSE and the modification notice, I think it's because of the date the LICENSE file and the NFT module is added? Should we align them?

tomtau:
it's probably ok as it is

PR #827, opened 2022-08-03
Problem: Simulation Tests fail on latest master

devashishdxt:
Solution: Changed from random genesis timestamp to current timestamp (wasmd depends on `UnixNano()` which can return undefined/negative value if chosen year is not in 1678-2262 https://github.com/CosmWasm/wasmd/blob/288609255ad92dfe5c54eae572fe7d6010e712eb/x/wasm/types/types.go#L261).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,7 +1,3 @@
 [submodule "cosmos-sdk"]
 	path = third_party/cosmos-sdk
 	url = https://github.com/cosmos/cosmos-sdk.git
-
-[submodule "wasmd"]
-	path = third_party/wasmd
-	url = https://github.com/CosmWasm/wasmd.git
--- a/app/sim_test.go
+++ b/app/sim_test.go
@@ -3,8 +3,10 @@ package app
 import (
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"math/rand"
 	"os"
+	"path/filepath"
 	"testing"
 
 	"github.com/stretchr/testify/require"
@@ -63,6 +65,30 @@ func interBlockCacheOpt() func(*baseapp.BaseApp) {
 	return baseapp.SetInterBlockCache(store.NewCommitKVStoreCacheManager())
 }
 
+// SetupSimulation wraps simapp.SetupSimulation in order to create any export directory if they do not exist yet
+func SetupSimulation(dirPrefix, dbName string) (simtypes.Config, dbm.DB, string, log.Logger, bool, error) {
+	config, db, dir, logger, skip, err := simapp.SetupSimulation(dirPrefix, dbName)
+	if err != nil {
+		return simtypes.Config{}, nil, "", nil, false, err
+	}
+
+	paths := []string{config.ExportParamsPath, config.ExportStatePath, config.ExportStatsPath}
+	for _, path := range paths {
+		if len(path) == 0 {
+			continue
+		}
+
+		path = filepath.Dir(path)
+		if _, err := os.Stat(path); os.IsNotExist(err) {
+			if err := os.MkdirAll(path, os.ModePerm); err != nil {
+				panic(err)
+			}
+		}
+	}
+
+	return config, db, dir, logger, skip, err
+}
+
 func TestFullAppSimulation(t *testing.T) {
 	config, db, dir, logger, skip, err := simapp.SetupSimulation("leveldb-app-sim", "Simulation")
 	if skip {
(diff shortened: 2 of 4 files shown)


tomtau (commented):
won't `time.now()` make it non-deterministic? `simtypes.RandTimestamp(r)` will have the same timestamp given the same seed

tomtau:
instead of the current time, the deterministic solution may be to constrain `simtypes.RandTimestamp(r)` to fall in that year range?

devashishdxt:
> instead of the current time, the deterministic solution may be to constrain `simtypes.RandTimestamp(r)` to fall in that year range?

Done

devashishdxt:
@tomtau @yihuang State determinism simulations pass if I use `fauxMerkleModeOpt` instead of `interBlockCacheOpt()` here: https://github.com/crypto-org-chain/chain-main/blob/680cc1cdb1e56a8405c2d36d5ca9bb7295d761bc/app/sim_test.go#L361. 

Same happens on `wasmd`. Do you know why this could happen?

tomtau:
@adu-crypto

adu-crypto:
> @tomtau @yihuang State determinism simulations pass if I use `fauxMerkleModeOpt` instead of `interBlockCacheOpt()` here:
> 
> https://github.com/crypto-org-chain/chain-main/blob/680cc1cdb1e56a8405c2d36d5ca9bb7295d761bc/app/sim_test.go#L361
> 
> Same happens on `wasmd`. Do you know why this could happen?

`fauxMerkleModeOpt ` means no real commitment of the merkle tree, which is only set for fast simulation.
So `fauxMerkleModeOpt` should not be used here I think.
I can help investigate the non determinism issue

devashishdxt:
> > @tomtau @yihuang State determinism simulations pass if I use `fauxMerkleModeOpt` instead of `interBlockCacheOpt()` here:
> > https://github.com/crypto-org-chain/chain-main/blob/680cc1cdb1e56a8405c2d36d5ca9bb7295d761bc/app/sim_test.go#L361
> > 
> > Same happens on `wasmd`. Do you know why this could happen?
> 
> `fauxMerkleModeOpt ` means no real commitment of the merkle tree, which is only set for fast simulation. So `fauxMerkleModeOpt` should not be used here I think. I can help investigate the non determinism issue

Just FYI, non-determinism is in `wasmd`. Same failure happens in `wasmd` as well and when I run simulations without including `wasmd`, it passes. Maybe some non-determinism included in sorting fix of wasmd?

adu-crypto:
@devashishdxt there is a backported fix for simulation determinism in release/v0.45.x: https://github.com/cosmos/cosmos-sdk/commit/5e40a1446056e862d3b32e56db6d93f5744056bc.
And this fix was included in 0.45.7, so after upgrading cosmos-sdk version to 0.45.7 in both wasmd and chain-main repo that I forked , `make test-sim-nondeterminism` passed successfully.

adu-crypto:
you could merge this: https://github.com/crypto-org-chain/wasmd/pull/3

devashishdxt:
> you could merge this: [crypto-org-chain/wasmd#3](https://github.com/crypto-org-chain/wasmd/pull/3)

I tested with this fix. The problem is not because of cosmos-sdk's non-determinism. There is some other source of non-determinism.

tomtau:
> `make test-sim-nondeterminism` passed successfully.
@adu-crypto was lucky?

adu-crypto:
It took a very long period that I thought it could pass while it finally failed with seed 2/3 in the 5th round.

> Simulating... block 48/100, operation 200/345. Logs to writing to /Users/adudu/.simapp/simulations/2022-08-05_12:21:08.log
    simulate.go:299: error on block  48/100, operation (250/345) from x/nft:
        failed to execute message; message index: 0: NFT wzn already exists in collection doggos: nft already exists [/Users/adudu/chain-main/x/nft/keeper/keeper.go:51]
        Comment: failed to execute message; message index: 0: NFT wzn already exists in collection doggos: nft already exists
--- FAIL: TestAppStateDeterminism (1599.42s)
FAIL
FAIL	github.com/crypto-org-chain/chain-main/v4/app	1602.147s
FAIL

but the failure is not owing to some non-determinism error but an error originated from `x/nft/simulation/operations.go`

adu-crypto:
I think the problem is that you are fetching an existing denomination for minting:
```go
func getRandomDenom(ctx sdk.Context, k keeper.Keeper, r *rand.Rand) string {
	var denoms = []string{kitties, doggos}
	i := r.Intn(len(denoms))
	return denoms[i]
}

func SimulateMsgMintNFT(k keeper.Keeper, ak types.AccountKeeper, bk types.BankKeeper) simtypes.Operation {
	return func(
		r *rand.Rand, app *baseapp.BaseApp, ctx sdk.Context, accs []simtypes.Account, chainID string,
	) (
		opMsg simtypes.OperationMsg, fOps []simtypes.FutureOperation, err error,
	) {
		denom, err := k.GetDenom(ctx, getRandomDenom(ctx, k, r))
...
```

devashishdxt:
Closing this as this is not relevant anymore.

PR #830, opened 2022-08-10
Problem: `wasmd` is not removed from `chain-main`

devashishdxt:
Solution: Remove `wasmd` + remove unncessary `replace` in `go.mod`


tomtau (approved):
removal seems ok, I haven't looked in detail into app.go whether it didn't reorder inits or make any mistakes... just need to note to verify those before final releases

PR #831, opened 2022-08-10
Problem: Cosmos SDK and ibc-go are not upgraded to `0.45.7` and `4.0.0-rc3`

devashishdxt:
Solution: Upgrade versions for both `cosmos-sdk` and `ibc-go`


app/test_helpers.go
```diff
@@ -330,6 +331,7 @@ func SignCheckDeliver(
 ) (sdk.GasInfo, *sdk.Result, error) {
 
 	tx, err := helpers.GenTx(
+		rand.New(rand.NewSource(time.Now().UnixNano())),
```
yihuang, line 334: probably need a "nolint" here

app/test_helpers.go
```diff
@@ -330,6 +331,7 @@ func SignCheckDeliver(
 ) (sdk.GasInfo, *sdk.Result, error) {
 
 	tx, err := helpers.GenTx(
```
tomtau, line 333: wouldn't `time.now()` not make it non-deterministic? or are those generated txs not regenerated in simulations?

devashishdxt: We only use this in (nft) unit tests.

tomtau (approved):

devashishdxt:
Forgot to update `gomod2nix`

PR #834, opened 2022-08-12
Problem: Fee middleware is not enabled for ICA module

devashishdxt:
Solution: Enable fee middleware for ICA module


tomtau (approved):
does it work now? there's a fail on the ibc integration test: https://github.com/crypto-org-chain/chain-main/runs/7798690698?check_suite_focus=true

devashishdxt:
> does it work now? there's a fail on the ibc integration test: https://github.com/crypto-org-chain/chain-main/runs/7798690698?check_suite_focus=true

Fixed app wiring. It should work now.

PR #841, opened 2022-08-16
Problem: `x/group` module not added and tested

devashishdxt:
Solution: Added `x/group` with integration tests

### DIFF
--- a/integration_tests/test_group.py
+++ b/integration_tests/test_group.py
@@ -1,0 +1,215 @@
+import json
+
+import pytest
+
+pytestmark = pytest.mark.normal
+
+
+def test_group(cluster, tmp_path):
+    cli = cluster.cosmos_cli()
+
+    admin = cluster.address("community")
+    signer1 = cluster.address("signer1")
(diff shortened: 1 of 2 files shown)


integration_tests/test_group.py
```diff
@@ -0,0 +1,215 @@
...
+            home=cli.data_dir,
+            node=cli.node_rpc,
+            keyring_backend="test",
+            chain_id=cli.chain_id,
+        )
+    )
+
+    assert rsp["code"] == 0, rsp["raw_log"]
+
+    # check if the proposal executed successfully
```
tomtau, line 204: two more assertion to check that the amount was transferred successfully (deducted from the group / sender, and added to the receiver)?

tomtau (approved):
is there some reference app for group wiring? simapp?

devashishdxt:
> is there some reference app for group wiring? simapp?

Yes. `simapp`.

integration_tests/test_group.py
devashishdxt: Done

PR #844, opened 2022-08-18
Problem: min commission is not set (fixes #843)

### LINKED ISSUE #843: Problem: min commission is not set
right now, the default value is zero in 0.46

tomtau:
Solution: added a custom upgrade handler to raise
the min commission to 5% to all validators after the upgrade is executed


tomtau:
may need to wait for this one to be fixed first: https://github.com/crypto-org-chain/chain-main/pull/842

PR #845, opened 2022-08-19
Problem: min commission is not set (fixes #843)

### LINKED ISSUE #843: Problem: min commission is not set

tomtau:
Solution: added a custom upgrade handler to raise
the min commission to 5% to all validators after the upgrade is executed

### DIFF
--- a/integration_tests/test_reward.py
+++ b/integration_tests/test_reward.py
@@ -12,7 +12,7 @@ def test_reward(cluster):
     validator1_address = cluster.address("validator", i=0)
     validator2_address = cluster.address("validator", i=1)
     # starts with crocncl1
-    validator1_operator_address = cluster.address("validator", i=0, bech="val")
+    # validator1_operator_address = cluster.address("validator", i=0, bech="val")
     validator2_operator_address = cluster.address("validator", i=1, bech="val")
     signer1_old_balance = cluster.balance(signer1_address)
     amount_to_send = 2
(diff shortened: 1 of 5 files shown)


devashishdxt:
Upgrade integration tests fail.

tomtau:
@devashishdxt it should be fixed now -- the new config option in pystarport's yaml got lost during a rebase

devashishdxt (changes requested):
There's a lint failure:

```
./integration_tests/test_reward.py:15:5: F841 local variable 'validator1_operator_address' is assigned to but never used
```

integration_tests/test_reward.py
```diff
@@ -19,7 +19,9 @@ def test_reward(cluster):
     fees = 3_000_000_000
     # wait for initial reward processed, so that distribution values can be read
     wait_for_block(cluster, 2)
-    old_commission_amount = cluster.distribution_commission(validator1_operator_address)
+    old_commission_amount = (
+        0  # cluster.distribution_commission(validator1_operator_address)
```
yihuang, line 23: this comment cause a linter issue. BTW, why comment out the call?

tomtau: because the config changed to have 0 commission -- that call failed due to that: https://github.com/crypto-org-chain/chain-main/runs/7912468478?check_suite_focus=true

I could possibly prepare a separate config for upgrade, but it seems simpler to just adjust this test

Terra854:
You might want to also bump the validators' commission rate to 5% during the upgrade too cause it appears that the individual validators' commission rate remains untouched

devashishdxt (approved):

PR #851, opened 2022-09-02
Problem: Protobuf code generation is not consistent with cosmos-sdk

devashishdxt:
Solution: Generate proto code using buf with similar configuration as cosmos-sdk


proto/buf.yaml
```diff
@@ -0,0 +1,19 @@
+# This module represents the "proto" root.
+version: v1
+name: buf.build/crypto-org-chain/chain-main
```
tomtau, line 3: this name/url gives 404 -- do we need to register it on buf.build?

tomtau (approved):
it seems the third party Git sub-modules are no longer needed after this (the needed protobuf definitions are fetched via buf?)? so could they be removed?

proto/buf.yaml
devashishdxt: https://docs.buf.build/bsr/usage#push-a-module

devashishdxt:
> it seems the third party Git sub-modules are no longer needed after this (the needed protobuf definitions are fetched via buf?)? so could they be removed?

They're still needed for swagger documentation.

proto/buf.yaml
tomtau: ok, for releases, we can perhaps specify an action to do it: https://docs.buf.build/ci-cd/github-actions

devashishdxt: Created #852

yihuang:
Shall we update the `config.json` too? similar to https://github.com/crypto-org-chain/cronos/pull/677/files#diff-b8ef6a5786bfd5df3147645f82844be943e1fe6d68de3b6b4681a941e077b8d7

devashishdxt:
> Shall we update the `config.json` too? similar to https://github.com/crypto-org-chain/cronos/pull/677/files#diff-b8ef6a5786bfd5df3147645f82844be943e1fe6d68de3b6b4681a941e077b8d7 The legacy APIs are removed.

Done

yihuang:
> > Shall we update the `config.json` too? similar to https://github.com/crypto-org-chain/cronos/pull/677/files#diff-b8ef6a5786bfd5df3147645f82844be943e1fe6d68de3b6b4681a941e077b8d7 The legacy APIs are removed.
> 
> Done

you need to regenerate the swagger to make it up to date.

devashishdxt:
> > > Shall we update the `config.json` too? similar to https://github.com/crypto-org-chain/cronos/pull/677/files#diff-b8ef6a5786bfd5df3147645f82844be943e1fe6d68de3b6b4681a941e077b8d7 The legacy APIs are removed.
> > 
> > 
> > Done
> 
> you need to regenerate the swagger to make it up to date.

done

yihuang:
> > > > Shall we update the `config.json` too? similar to https://github.com/crypto-org-chain/cronos/pull/677/files#diff-b8ef6a5786bfd5df3147645f82844be943e1fe6d68de3b6b4681a941e077b8d7 The legacy APIs are removed.
> > > 
> > > 
> > > Done
> > 
> > 
> > you need to regenerate the swagger to make it up to date.
> 
> done

and statik needs to be updated too ;D

devashishdxt:
> and statik needs to be updated too ;D

Done

yihuang (approved):
LGTM

PR #863, opened 2022-09-23
Problem: gosec fails to upload analysis results

tomtau:
Solution: updated the gh action permissions


devashishdxt:
Is there a way to exclude `third_party` git submodules from this analysis?

tomtau:
@devashishdxt I removed the submodule checkout, so they may be empty in that workflow, we'll see if it works

devashishdxt (approved):

PR #869, opened 2022-09-29
Problem: ibc-go is not upgraded to v5.0.0

devashishdxt:
Solution: Upgrade ibc-go to v5.0.0


tomtau (approved):
there's also a lot of patches on 0.46.x branch of SDK, but I guess we'll want to update after there's a 0.46.2 (given the current one still doesn't have the gRPC nil fix)

PR #870, opened 2022-09-29
Problem: state streaming is not supported in old versions

tomtau:
Solution: backported state streaming (+ grpc nil fix) to 0.44.x fork basic integration added based on https://github.com/crypto-org-chain/cronos/pull/717

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -6,10 +6,12 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"sync"
 	"time"
 
 	"github.com/gorilla/mux"
 	"github.com/rakyll/statik/fs"
+	"github.com/spf13/cast"
 
 	abci "github.com/tendermint/tendermint/abci/types"
 	tmjson "github.com/tendermint/tendermint/libs/json"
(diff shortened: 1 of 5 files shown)


tomtau:
@yihuang https://github.com/crypto-org-chain/cosmos-sdk/compare/v0.44.8-patch3...v0.44.8-patch4 -- are there other patches needed (I saw some fixes for handler closing etc.)? and is the config extension needed?

yihuang:
> I saw some fixes for handler closing etc

do you mean the fix for the `--grpc-only` mode? only need if we'll use this mode.

tomtau:
that one, plus https://github.com/cosmos/cosmos-sdk/pull/13394/files or others... but not sure how much gprc-only mode is used

yihuang:
> that one, plus https://github.com/cosmos/cosmos-sdk/pull/13394/files or others... but not sure how much gprc-only mode is used

I guess not many people use it right now, but it could be useful in the distributed query service.

tomtau:
moved to https://github.com/crypto-org-chain/chain-main/pull/881

PR #874, opened 2022-09-29
Problem: potential resource leaks are not checked

tomtau:
Solution: added a CI static analysis for resource leaks


yihuang (approved):
LGTM

PR #882, opened 2022-10-19
Problem: `go` version is not upgraded to latest

devashishdxt:
Solution: Upgraded `go` version to `1.19.2`

PR #883, opened 2022-10-21
Problem: Cosmos SDK not upgraded to include patch for dragonberry

devashishdxt:
Solution: Upgrade cosmos-sdk to 0.46.3 + update submodules


tomtau (approved):

devashishdxt:
Vulnerability checks will be fixed once we upgrade go to `v1.19.2` (#882).

tomtau:
@yihuang GH interfaces doesn't show it on PR, but the action was executed and there's still an error: https://github.com/crypto-org-chain/chain-main/actions/runs/3319368641/jobs/5484441848#step:5:2754

tomtau:
good to bump to 0.46.4 with that upgrade interruption patch: https://github.com/crypto-org-chain/cronos/pull/748/files

yihuang:
> https://github.com/crypto-org-chain/chain-main/actions/runs/3319368641/jobs/5484441848#step:5:2754

strange, why it don't run in CI.
It runs successfully locally though, `nix-shell integration_tests/shell.nix --run "pytest -s integration_tests/test_multisig.py"`

tomtau:
@yihuang I restarted it in https://github.com/crypto-org-chain/chain-main/pull/884 and it doesn't have that error for multisig

but still an error for IBC (ica auth
https://github.com/crypto-org-chain/chain-main/actions/runs/3325759773/jobs/5498781620#step:5:2575 ) + upgrade

tomtau:
this now seems to pass all integration tests: https://github.com/crypto-org-chain/chain-main/pull/884 but not sure if the approach with cosmovisor is correct

PR #900, opened 2022-11-23
Problem: semgrep results are not uploaded

tomtau:
Solution: add the app token


tomtau (commented):
will merge, as it's already on the cronos repo

PR #903, opened 2022-11-29
Problem: `query nft owner` returns error when user has NFTs transferred over IBC

### LINKED ISSUE #901: Problem: `query nft owner` returns error when user has NFTs transferred over IBC
`query nft owner [address] --denom-id=<denom-id>` returns an error when use has NFTs transferred over IBC. This error happens only when user does not explicitly provide `denom-id`.

devashishdxt:
Solution: Fixed denomID and tokenID splitting logic when denom has `ibc`. Fixes #901.

### DIFF
--- a/x/nft/types/keys.go
+++ b/x/nft/types/keys.go
@@ -58,12 +58,28 @@ func SplitKeyOwner(key []byte) (address sdk.AccAddress, denomID, tokenID string,
 
 func SplitKeyDenom(key []byte) (denomID, tokenID string, err error) {
 	keys := bytes.Split(key, delimiter)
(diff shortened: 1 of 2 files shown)


tomtau (approved):
`SplitKeyDenom` is only used in gRPC (`Owner`) -- is that via any path used on the consensus-related logic, i.e. is this a breaking change?

devashishdxt:
> `SplitKeyDenom` is only used in gRPC (`Owner`) -- is that via any path used on the consensus-related logic, i.e. is this a breaking change?

This is only used in `gRPC` query. This is not a breaking change.

PR #905, opened 2022-12-05
Problem: missing ICA messages in the upgrade plan

tomtau:
Solution: added the missing staking messages and ICS721 to the allowlist, plus changed the ugprade plan name


leejw51crypto (approved):
lgtm

PR #919, opened 2023-02-06
Problem: incorrect codecov reporting (fixes #278)

### LINKED ISSUE #278: Problem: incorrect codecov reporting?

tomtau:
Solution: switch to Go 1.20's binary coverage instrumentation


integration_tests/utils.py
```diff
@@ -143,19 +142,6 @@ def cluster_fixture(
         if post_init:
             post_init(chain_id, chain_data)
 
-        if enable_cov:
```
yihuang, line 146: We can remove the parameter as well.

yihuang (approved):

devashishdxt:
Is it still not reporting exact coverage from integration tests?

devashishdxt (approved):

tomtau:
@devashishdxt it's not visible in the main summary, but in "indirect" changes, it looks more correct: https://app.codecov.io/gh/crypto-org-chain/chain-main/pull/919/indirect-changes

PR #922, opened 2023-02-09
Problem: incorrect codecov reporting (fixes #278)

### LINKED ISSUE #278: Problem: incorrect codecov reporting?

tomtau:
Solution: add ignore for irrelevant paths; upload results from more integration tests

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,6 +1,7 @@
 name: nix flow
 on:
   pull_request:
+  merge_group:
   push:
     branches:
       - master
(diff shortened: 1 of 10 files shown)


yihuang (approved):

.github/workflows/nix.yml
```diff
@@ -38,12 +39,10 @@ jobs:
...
       - name: Run integration tests
         run: make nix-integration-test
-      - name: Set up Go
-        uses: actions/setup-go@v2
-        with:
-          go-version: 1.20.0
       - name: Convert coverage data
-        run: go tool covdata textfmt -i=coverage -o profile.txt
+        run: |
+           nix profile install nixpkgs#go_1_20
```
yihuang, line 44: you might need to specify `nixpkgs` channel in `cachix/install-nix-action` to do this, not sure if it can run.

yihuang: oh, I guess it'll use the `flake.lock` in project directory?

PR #949, opened 2023-05-03
Update bufbuild plugin references

mfridman:
This PR updates buf.gen.yaml to reference Buf-managed plugins, this is to ensure [upcoming Remote Package changes](https://buf.build/blog/remote-packages-remote-plugins-approaching-v1) do not break your builds.

I didn't spend much time getting `make` to build locally, so you may need to re-generate code. The updated plugins are found at:

- https://buf.build/protocolbuffers/python
- https://buf.build/grpc/python

I've also updated them to the latest available version, but that may not be what you want.

The source of theses packages plugins can be found at https://github.com/bufbuild/plugins


pystarport/buf.gen.yaml
```diff
@@ -2,7 +2,7 @@ version: v1
 managed:
   enabled: true
 plugins:
-  - remote: buf.build/protocolbuffers/plugins/python:v3.19.1-1
+  - plugin: buf.build/protocolbuffers/python:v22.3
```
mfridman, line 5: Newer versions of the `buf` CLI use the `plugin` key for both local and remote plugin references.

mmsqe (approved):

yihuang (approved):
The python protobuf files are generated on the fly, so I guess it's alright since the integration tests all passed.

yihuang:
> The base branch requires all commits to be signed

@mfridman  can you sign the commits?

Dubersnack (approved):

PR #1077, opened 2024-08-28
Changed Crypto.org to Cronos

kaz-CDC:
Changed all References of Crypto.org to Cronos.org

References in code blocks and within URLs e.g. "github.com/crypto-org-chain/chain-main/v4/x/chainmain" wasn't changed


mmsqe:
Hi @kaz-CDC, can you solve the conflict?

